PR #15, opened 2019-10-02
cmake: Fix how compile definition is added

SebastianBoe:
The compile definition ENET_ENHANCEDBUFFERDESCRIPTOR_MODE has been
added by modifying the build environment of the 'zephyr' library. But
modifying this library is not supported.

The intention was to modify the global build environment, so we change
to use the zephyr_ API instead of the zephyr_library_ API as this API
will behave as intended.

Signed-off-by: Sebastian Bøe <sebastian.boe@nordicsemi.no>

### DIFF
--- a/mcux/drivers/kinetis/CMakeLists.txt
+++ b/mcux/drivers/kinetis/CMakeLists.txt
@@ -6,8 +6,9 @@
 
 zephyr_include_directories(.)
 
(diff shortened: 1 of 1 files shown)


SebastianBoe:
@MaureenHelm , @galak : Review needed

PR #24, opened 2019-12-13
components: add power down register mask

agansari:
Add a mask for KSZ8081 PHY's control register to set power down.

Signed-off-by: Andrei Gansari <andrei.gansari@nxp.com>


agansari:
all checks passed in https://github.com/zephyrproject-rtos/zephyr/pull/21370
shippable https://app.shippable.com/github/zephyrproject-rtos/zephyr/runs/58396/summary

@MaureenHelm @galak please merge

PR #36, opened 2020-03-30
mcux: drivers: imx: fix spi timing delay overflow

lgl88911:
define realDelay by uint32_t may be overflow.
overflow will cause delay cal error.

Signed-off-by: Frank Li <lgl88911@163.com>

### DIFF
--- a/mcux/README
+++ b/mcux/README
@@ -95,3 +95,6 @@ Patch List:
    * Add USB device controller drivers, the drivers are based on MCUXpresso SDK release tag REL_2.5.0_REL9_RFP_RC3_7_1.
 
    * Add EDMA bug fix for EDMA_SubmitTransfer busy state judgement, this fix applys to 2.7.0
+
+   * Add Time delay bug fix for LPSPI_MasterSetDelayTimes of fsl_lpspi.c to avoid realDelay and bestDelay overflow.
+
--- a/mcux/drivers/imx/fsl_lpspi.c
+++ b/mcux/drivers/imx/fsl_lpspi.c
@@ -577,7 +577,7 @@ uint32_t LPSPI_MasterSetDelayTimes(LPSPI_Type *base,
                                    lpspi_delay_type_t whichDelay,
                                    uint32_t srcClock_Hz)
 {
-    uint32_t realDelay, bestDelay;
+    uint64_t realDelay, bestDelay;
     uint32_t scaler, bestScaler;
     uint32_t diff, min_diff;
     uint64_t initialDelayNanoSec;


lgl88911:
I can't confirm whether hal_nxp of module accepts the modification of the driver code. If so, I will go to zephyr  submit a pr to verify this pr.

MaureenHelm:
With what arguments do you get an overflow?

lgl88911:
> With what arguments do you get an overflow?

uint32_t realDelay,bestDelay  max is 4294967295
The following code is included in LPSPI_MasterSetDelayTimes
bestDelay will be set error value by overflow, 257U*1000000000U>4294967295
when (scaler + 1U + additionalScaler)>4 , realDelay will overflow.

```
        bestDelay = 1000000000U;
        bestDelay *= 257U; /* based on DBT+2, or 255 + 2 */ 
        bestDelay /= clockDividedPrescaler;

...

    for (scaler = 0U; scaler < 256U; scaler++)
    {
        if (min_diff == 0U)
        {
            break;
        }
        /* Calculate the real delay value as we cycle through the scaler values.
        Due to large size of calculated values (uint64_t), we need to break up the
        calculation into several steps to ensure accurate calculated results
        */
        realDelay = 1000000000U;
        realDelay *= (scaler + 1U + additionalScaler);
        realDelay /= clockDividedPrescaler;
```

mcux/drivers/imx/fsl_lpspi.c
```diff
@@ -1198,6 +1198,7 @@ static void LPSPI_MasterTransferComplete(LPSPI_Type *base, lpspi_master_handle_t
 {
     assert(handle);
 
+    LPSPI_ClearStatusFlags(base, kLPSPI_TransferCompleteFlag);
```
MaureenHelm, line 1201: I'm not sure this is the right fix. Can you look at updating `spi_mcux_master_transfer_callback` to parse the `status` argument instead?

If it is the right fix, it should be in a separate commit from the uint32_t overflow change.

lgl88911: Thanks for your review.
clean flag can be put in spi_mcux_master_transfer_callback, when status == kStatus_Success. But, I saw in LPSPI_SlaveTransferHandleIRQ in HAL NXP that the clear interrupt flag is being processed. I think that the clear interrupt flag processing of Master transmission should also be put in HAL NXP.

lgl88911: In this PR, the interrupted modification is removed and only the problem of time delay is dealt with. When the interrupted modification method is confirmed, I submit another commit.

MaureenHelm: Thank you. Will you please also update the patch list in `mcux/README` to mention the delay change, and then I'll approve this PR.

lgl88911: I have update the patch list, Please review again, Thanks

mcux/README
```diff
@@ -92,3 +92,5 @@ Patch List:
      support is enabled.
 
    * Add USB device controller drivers, the drivers are based on MCUXpresso SDK release tag REL_2.5.0_REL9_RFP_RC3_7_1.
+
+   * Change 32 to 64 to avoid time delay calculation error caused by realDelay and bestDelay overflow.
```
MaureenHelm, line 96: Need to mention the fsl_lpspi driver

lgl88911: Thanks for your patient guidance, I have performed these steps, please review again

henrikbrixandersen:
This issue is also fixed by upstream SDK v2.8 in #56.

lgl88911:
@henrikbrixandersen， Thanks for your infomation, https://github.com/zephyrproject-rtos/hal_nxp/pull/56 has bee merged, so close this PR

PR #44, opened 2020-05-28
mcux: drivers: imx: fix Other interrupts treated as transfer completion

lgl88911:
After the master transfer is completed,
if the sr flag is not clear, the callback of the transfer
completion will be called again in other interrupts.

Signed-off-by: Frank Li <lgl88911@163.com>

### DIFF
--- a/mcux/drivers/imx/fsl_lpspi.c
+++ b/mcux/drivers/imx/fsl_lpspi.c
@@ -1198,6 +1198,8 @@ static void LPSPI_MasterTransferComplete(LPSPI_Type *base, lpspi_master_handle_t
 {
     assert(handle);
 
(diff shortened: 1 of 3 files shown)


mcux/drivers/imx/fsl_lpspi.c
```diff
@@ -1198,6 +1198,8 @@ static void LPSPI_MasterTransferComplete(LPSPI_Type *base, lpspi_master_handle_t
 {
     assert(handle);
 
+    LPSPI_ClearStatusFlags(base, kLPSPI_TransferCompleteFlag);
```
MaureenHelm, line 1201: Will you please make the same fix in `mcux/drivers/kinetis/fsl_lpspi.c`

MaureenHelm: thank you!

PR #49, opened 2020-07-09
Add support for MXRT600

mmahadevan108:
The code currently can successfully run samples/hello_world & samples/basic/blinky


mmahadevan108:
cc @MaureenHelm

mmahadevan108:
> So one change instead of mcux/devices/MIMXRT685S can this be more of a family so one could support both MIMXRT685S and MIMXRT633S w/o having to do a full import of all the driver files.

All drivers files are kept out of the device folder and can be shared between MIMXRT685S & MIMXRT633S. When adding MIMXRT633S, we should be able add a folder for MIMXRT633S under devices.

galak:
> All drivers files are kept out of the device folder and can be shared between MIMXRT685S & MIMXRT633S. When adding MIMXRT633S, we should be able add a folder for MIMXRT633S under devices.

My point is that duplicating this for every chip is not scalable.  Thus suggesting that the dir with actual driver code be for the RT6xx family.

mmahadevan108:
> > All drivers files are kept out of the device folder and can be shared between MIMXRT685S & MIMXRT633S. When adding MIMXRT633S, we should be able add a folder for MIMXRT633S under devices.
> 
> My point is that duplicating this for every chip is not scalable. Thus suggesting that the dir with actual driver code be for the RT6xx family.
The directory drivers/imxrtxxx is meant for the all RT6XX family, I can rename this to be more clear i.e drivers/imxrt6xx. 
We are in the process of restructuring the drivers folder under mcux to avoid such duplication.

galak:
I'd suggest re-ordering the commits so that the file import is first, than the cmake additions.  Also, squash the removal of the power libs into the file import commit.

mcux/drivers/CMakeLists.txt
```diff
@@ -1,10 +1,12 @@
 #
-# Copyright (c) 2018, NXP
+# Copyright (c) 2018-2020, NXP
 #
 # SPDX-License-Identifier: Apache-2.0
 #
 
-if("${CONFIG_SOC_FAMILY}" STREQUAL "nxp_imx")
+if("${CONFIG_SOC}" STREQUAL "mimxrt685s_cm33")
```
galak, line 7: Can we change this from CONFIG_SOC to based on something like SOC_SERIES?

galak:
Also rename the commit message for the file import

mcux/drivers/CMakeLists.txt
mmahadevan108: Done

mmahadevan108:
> I'd suggest re-ordering the commits so that the file import is first, than the cmake additions. Also, squash the removal of the power libs into the file import commit.

Done

mmahadevan108:
> Also rename the commit message for the file import

Done

mmahadevan108:
> LGTM. Probably good to update mcux/README

Done.

galak:
looks like you need a rebase to deal with merge conflict on mcux/README.

mmahadevan108:
Rebase done

PR #64, opened 2020-10-29
iMX: can: Fix building flexcan driver

galak:
We need to same ASSERT/NDEBUG support that exists for kinetis for imx.
Otherwise we get build errors with building the flexcan drivers on
imx-rt.

Signed-off-by: Kumar Gala <kumar.gala@linaro.org>


galak:
Seeing failures in nightly when building `samples/drivers/can` on mimxrt1060_evk or mimxrt1064_evk

henrikbrixandersen (approved):

MaureenHelm (approved):
Thanks. @henrikbrixandersen even asked about this when we added extended flexcan driver support from kinetis to imx. I didn't see this warning because I was building only `tests/subsys/canbus/isotp/conformance`, not `samples/drivers/can`.

PR #70, opened 2020-12-04
enet: dsa: Extend the ENET_FRAME_MAX_FRAMELEN to accommodate 1B DSA tag

lmajewski:
The ENET_FRAME_MAX_FRAMELEN needs to be adjusted to allow reading extra
bytes (i.e. tags) appended by DSA switches (like KSZ8794) to the received
frame.

Without this change - frames with maximal length (i.e. 1588) will be
dropped as tag bytes (1 byte in case of KSZ8794) will be truncated and
not provided to upper Zephyr layers.

Signed-off-by: Lukasz Majewski <lukma@denx.de>

---
I do have some questions regarding this patch:
1. It looks like I would need some conditional alteration of `ENET_FRAME_MAX_FRAMELEN` to avoid side effects when 
the DSA is not supported/enabled (https://github.com/zephyrproject-rtos/zephyr/pull/27620). I'm wondering how I can provide it - as hal_nxp should to be a separate layer (on top of which Zephyr runs). 

I could do (in hal/nxp/mcux/drivers/kinetis/fsl_enet.h):
`#if defined(CONFIG_NET_DSA) && defined(CONFIG_DSA_KSZ8794_TAIL_TAGGING)`
`#define ENET_FRAME_MAX_FRAMELEN 1519`
`#endif`

But I'm not sure if this is the correct approach. To be more specific - I'm afraid of Zephyr layers violation.

2. The other idea would be to add new board like **ip_k66f** ,which do have the switch onboard. However, it can work with and without tagging enabled (i.e. the KSZ8794 switch can be configured to add the tag or not), so I don't see how this can be done.
Moreover, adding a separate board seems like an overkill to just modify the single define.

Ideas and help more than welcome.

### DIFF
--- a/mcux/drivers/kinetis/fsl_enet.h
+++ b/mcux/drivers/kinetis/fsl_enet.h
@@ -118,7 +118,7 @@
 
 /*! @name Defines some Ethernet parameters. */
 /*@{*/
-#define ENET_FRAME_MAX_FRAMELEN 1518U /*!< Default maximum Ethernet frame size. */
+#define ENET_FRAME_MAX_FRAMELEN 1519U /*!< Default maximum Ethernet frame size. */
 
 #define ENET_FIFO_MIN_RX_FULL  5U                                        /*!< ENET minimum receive FIFO full. */
 #define ENET_RX_MIN_BUFFERSIZE 256U                                      /*!< ENET minimum buffer size. */


lmajewski:
Can we have discussion regarding this issue?

The DSA PR has been now merged, so it would be more than welcome to have this sorted out.

mmahadevan108:
@lmajewski Sorry for the delay in getting back. The NXP SDK team is updating the ENET driver along the below lines
_________________
I didn't extend frame length macro. According to my research, DSA tag should share similar arch with VLAN. If user needs to use some switch vendor tag, they can enable VLAN when init, driver will extend frame length as 4 or 8(SVLAN enable).
Or if there's demand to use VLAN+DSA together, set the config->rxMaxFrameLen to a higher value, this is a more flexible way for all situation, even send longer jumbo frame. ENET_FRAME_MAX_FRAMELEN is an reference value for default case, application also need to extend the Rx buffer if VLAN/SVLAN enable or config->rxMaxFrameLen takes a longer value.
___________________

Once this patch is finalized, we will update the hal layer. Hope this works.

lmajewski:
@mmahadevan108 
I'm not sure if I've understood the explanation / idea behind this issue.

Please correct me if I'm wrong, but it seems like the VLAN approach is a bit odd for me.
Do you suggest to enable in NXP HAL the `VLAN` frames support (even when we will not use VLAN) so extra 4 bytes will be passed from this HW layer to Zephyr?
Could you elaborate more on this idea?

The issue is that different DSA switches (KSZ8794 in our case) can have different size of tags. It can be 1B, 2B, 4B. The 
problem will be evident when one wants to pass VLAN frame from SoC (NXP) with appended tag. The last byte (with tag) will be just discarded when we pass frame from Zephyr to NXP HAL layer (as ENET_FRAME_MAX_FRAMELEN is not big enough to accommodate tag).

I rather thought about:
1. Some convenient define to allow appending some arbitrary bytes to frames (as mentioned above in this discussion)
or
2. Some script invoked when `west` builds Zephyr so `#defines` are properly adjusted when DSA is supported.

mmahadevan108:
The suggestion is to use config->rxMaxFrameLen. This is set in the ENET_RCR register and will be read when doing the frame length checks.

lmajewski:
@mmahadevan108 
Please correct me if I've misunderstood something. The idea is as follows:
1. In the `modules/hal/nxp/mcux/drivers/kinetis/fsl_enet.c` there are some places where `ENET_FRAME_MAX_FRAMELEN`
is used directly to check the length of the frame - like in `ENET_SendFrame()`. This macro shall be replaced with `config->rxMaxFrameLen`.
2. In Zephyr's - `modules/hal/nxp/mcux/drivers/kinetis/eth_mcux.c` `eth_mcux_init()` then I could adjust `config->rxMaxFrameLen` to be 1519 bytes (instead of 1518).

Am I correct?

And a side question - do you know when the proper patch for NXP_HAL will be available?

mmahadevan108:
> 1. In the `modules/hal/nxp/mcux/drivers/kinetis/fsl_enet.c` there are some places where `ENET_FRAME_MAX_FRAMELEN`
>    is used directly to check the length of the frame - like in `ENET_SendFrame()`. This macro shall be replaced with `config->rxMaxFrameLen`.

The checks for length will look at the `MAX_FL `bits in the `ENET_RCR `register.

> 2. In Zephyr's - `modules/hal/nxp/mcux/drivers/kinetis/eth_mcux.c` `eth_mcux_init()` then I could adjust `config->rxMaxFrameLen` to be 1519 bytes (instead of 1518).
> 
> Am I correct?

Thats correct.


> And a side question - do you know when the proper patch for NXP_HAL will be available?

I am checking with the team on the date for a release with this fix, I will keep you posted.

lmajewski:
This patch is not needed anymore as in `nxp/mcux/drivers/kinetis/fsl_enet.c` :
1. Afther the SHA1: 2740046f3fd5ff076c46b6f2c9f6790a51a222dd `Update frdm_k64f files to SDK 2.10.0`  the `ENET_FRAME_MAX_FRAMELEN` is not checked explicitly; instead macro, which reads MAX_FL from RCR register is used. This is not a problem as one passes length = 1501 to `ENET_SendFrame()` which is `< 1518` assigned by default.
2. Then the `(handle->txBuffSizeAlign[ringId] >= length)` is checked without issues as `txBuffSizeAlign` is set to 1520 (in `zephyr/drivers/ethernet/eth_mcux.c) after rounding up the buffer size. Those two extra bytes allow storing 1B tag from KSZ8794.  Devices with tags larger than 2B will encounter problems here.
3. The K66F's ENET HW will have the `MAX_FL` field of `RCR` register set to default value of 1518. This will not prevent this IP block from sending/receiving data. It will only set the `BABR` and `BABT` flags in `ENET_EIR` register. As corresponding interrupts in `EIMR` registers are masked - the transmission is performed as it would have 1518. not 1519 bytes.

**To sum up**
Considering the above - this pull request can be safely closed without taking any action.

PR #78, opened 2021-02-12
mcux: Add configs to make building drivers in hal_nxp optional

MaureenHelm:
In preparation for switching over to a new GitHub-based distribution of
the MCUXpresso SDK and a corresponding new module in the west manifest,
add options to configure which bare metal drivers in the hal_nxp module
to build. This will allow us support both the existing module and the
new module simultaneously until we are ready to remove the old one.

Prior to this change, the hal_nxp module had references to Zephyr driver
Kconfig symbols like CONFIG_ADC_MCUX_ADC12 and CONFIG_I2C_MCUX_LPI2C,
which made a two-way coupling between the hal_nxp module and the Zephyr
drivers. This two-way coupling made it difficult to introduce the new
GitHub-based MCUXpresso SDK module because we end up building the same
bare metal driver twice and then fail to link due to duplicate symbols.

This patch introduces a different approach to building the bare metal
drivers in hal_nxp and simplifies the coupling to go only one-way. Now,
the Zephyr drivers use Kconfig "imply" statements to enable building
their associated bare metal drivers in the hal_nxp module. The hal_nxp
module no longer needs to know about any Zephyr drivers.

This approach will also be used to build drivers in the new module. By
using "imply" instead of "select" we will be able to configure which
version of each bare metal driver to use.

Signed-off-by: Maureen Helm <maureen.helm@nxp.com>


hakehuang (approved):
do we have a schedule on when will the new hal repo phase in?

mmahadevan108 (approved):

MaureenHelm:
> do we have a schedule on when will the new hal repo phase in?

I'm aiming for the 2.6 release. I have a prototype working for a few drivers on `frdm_k64f` and hope to post a draft PR in the next week.

henrikbrixandersen (approved):

dleach02:
@MaureenHelm, in an effort to clean things up, I would like to close this PR if you approve

PR #81, opened 2021-03-18
MXRT600 Add OS Timer

mmahadevan108:


galak (commented):
Where is SystemCoreClock used?

mmahadevan108:
https://github.com/NXPmicro/zephyr/blob/Update_MXRT600_Timer/arch/arm/core/aarch32/cortex_m/timing.c#L38

I am working to switch RT600 to use a different timer instead of SYSTICK. This is done so we can start implementing SoC power management states on NXP platforms. Eventually this will grow to other NXP SoC's as well, hence this change.

mmahadevan108:
I have changed the PR to add system.c file only for RT600.

galak (approved):

galak:
At some point we should see if we define `SystemCoreClock` based on devicetree in the future.

MaureenHelm (approved):

PR #85, opened 2021-04-06
Add dtsi pin data for kinetis SoCs

galak:


galak:
Pulling comments from https://github.com/zephyrproject-rtos/zephyr/pull/33959 that need to be addressed from @henrikbrixandersen:

Just picking a random pin/function for this comment: Should the function here be adc0_se2_pta7? Otherwise, how would you know which ADC0 single-ended input, you are mapping?

From a quick glance, this comments seems to affect all "numbered" peripheral pins (e.g. cmp1_in_pta7 => cmp1_in1_pta7, ftm0_flt_pta7 => ftm0_flt2_pta7, ...).

galak:
> * Just picking a random pin/function for this comment: Should the function here be `adc0_se2_pta7`? Otherwise, how would you know which ADC0 single-ended input, you are mapping?
> 
> * From a quick glance, this comments seems to affect all "numbered" peripheral pins (e.g. `cmp1_in_pta7` => `cmp1_in1_pta7`, `ftm0_flt_pta7` => `ftm0_flt2_pta7`, ...).

These should now be resolved.

henrikbrixandersen (commented):
Looks great! One question, though. Does the repeated `/* < PIN PCR[MUX] > */` line provide any value? Loosing it could save quite a lot of lines...

galak:
> Looks great! One question, though. Does the repeated `/* < PIN PCR[MUX] > */` line provide any value? Loosing it could save quite a lot of lines...

Yeah, I'll drop/rework that so there is just one header block at the start of the file.

MaureenHelm (approved):

henrikbrixandersen (approved):
Thanks!

PR #93, opened 2021-07-05
hal: pxp: add support for the pxp of rt1170

CristXu:
add SDK support for the pxp

Signed-off-by: Crist Xu <crist.xu@nxp.com>


dleach02 (approved):

dleach02:
@CristXu I'm going to close this one for now. When we need the pxp driver we can reenable or do a new PR.

PR #117, opened 2021-09-15
Add PMIC pca9420 I2C driver to components directory

danieldegrasse:
This PR adds the PCA9420 PMIC driver from the RT600 evaluation board SDK to the mcux components directory. This driver is required to interface with the PCA9420 PMIC, which is necessary to correctly support interfacing with SD cards at 1.8V signalling levels on the RT600 evaluation board. The PMIC must be interfaced with in order to drop the peripheral voltage from 3.3 to 1.8V when using an SD card that supports this lower voltage

### DIFF
--- a/mcux/components/CMakeLists.txt
+++ b/mcux/components/CMakeLists.txt
@@ -6,4 +6,11 @@ if(CONFIG_ETH_MCUX)
   zephyr_include_directories(phyksz8081)
 endif()
 
+if(CONFIG_USE_PMIC_PCA9420)
+  zephyr_include_directories(pca9420)
+endif()
+
 zephyr_library_sources_ifdef(CONFIG_ETH_MCUX phyksz8081/fsl_phy.c)
+
+zephyr_library_sources_ifdef(CONFIG_SDMMC_USDHC_PMIC_1_8_VOL 
+                             pca9420/fsl_pca9420.c)
--- a/mcux/components/pca9420/fsl_pca9420.c
+++ b/mcux/components/pca9420/fsl_pca9420.c
@@ -1,0 +1,542 @@
+/*
+ * Copyright 2018-2020 NXP
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include "fsl_common.h"
+#include "fsl_pca9420.h"
+#include "fsl_power.h"
+
+/*******************************************************************************
+ * Definitions
+ ******************************************************************************/
+
+/*******************************************************************************
+ * Prototypes
+ ******************************************************************************/
+
+/*******************************************************************************
+ * Variables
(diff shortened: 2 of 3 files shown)


mcux/components/CMakeLists.txt
```diff
@@ -6,4 +6,11 @@ if(CONFIG_ETH_MCUX)
   zephyr_include_directories(phyksz8081)
 endif()
 
+if(CONFIG_SDMMC_USDHC_PMIC_1_8_VOL)
```
mmahadevan108, line 9: Can we add a more generic config to include the PMIC PCA9420 driver. This is in case future Zephyr drivers need this component. One suggestion for a name is `CONFIG_USE_PMIC_PCA9420`

galak (changes requested):
Why is this being added to the NXP HAL repo instead of to Zephyr?

danieldegrasse:
@galak it could be added to Zephyr directly, but I'm not sure where it would fit. The PMIC is being used as a regulator, but the specific use case is lowering its output voltage. Since Zephyr only supports regulators as an on/off device, there didn't seem to be a clear place to put it. To be clear, I'm open to and would prefer to integrate it into Zephyr, I just didn't see a way to do so cleanly.

galak:
> @galak it could be added to Zephyr directly, but I'm not sure where it would fit. The PMIC is being used as a regulator, but the specific use case is lowering its output voltage. Since Zephyr only supports regulators as an on/off device, there didn't seem to be a clear place to put it. To be clear, I'm open to and would prefer to integrate it into Zephyr, I just didn't see a way to do so cleanly.

I would say do it as a regulator and just have a custom API for the additional functionality at this point.

danieldegrasse:
> I would say do it as a regulator and just have a custom API for the additional functionality at this point.

Is there an example of where a custom API has been added in Zephyr previously I could base this code off of? Also, I'm aware this is a much longer term discussion, but I think there may be reason to allow setting voltage levels within the Zephyr regulator api, based on the [linux regulator api](https://www.kernel.org/doc/html/v5.0/driver-api/regulator.html#c.regulator_set_voltage)

galak:
> Is there an example of where a custom API has been added in Zephyr previously I could base this code off of? Also, I'm aware this is a much longer term discussion, but I think there may be reason to allow setting voltage levels within the Zephyr regulator api, based on the [linux regulator api](https://www.kernel.org/doc/html/v5.0/driver-api/regulator.html#c.regulator_set_voltage)

If you want to look at extending the zephyr regulator api that would be awesome.

Look at include/drivers/rtc/maxim_ds3231.h for an example of extending a driver w/a custom API.

danieldegrasse:
Closing this PR, as https://github.com/zephyrproject-rtos/zephyr/pull/38835 replaces it with native functionality in Zephyr

PR #119, opened 2021-10-04
drivers: [lpi2c]Fix issue of ignoring the nack signal by mistake

pdgendt:
Port [mcux-sdk fix](https://github.com/NXPmicro/mcux-sdk/commit/e84ccf99a203bbbd8872d9953c8d3b57864d484b) to zephyr, original message:

Issue: Customer reports that if only transmitting the slave address of non-existed slave on bus, no nack is returned.
Analysis: When transmitted data count is not 0, the last byte's nack is expected to be ignored. The checking logic does not consider transferring only the slave address, which makes the transferred count to be only 1, so the nack is ignored by mistake.
Resolution:
1.When the data size is zero, the nack detected during slave probing(including slave address and register address) cannot be ignored
2.After pushing all data to FIFO and nack is detected, only when the txfifo is empty, or have one byte which is the stop command, then the nack status can be ignored


pdgendt:
@mmahadevan108 any comments on this PR?

PR #120, opened 2021-10-04
Add RT1160 Support Files

danieldegrasse:
This PR adds the support files required to enable the RT1160 in Zephyr, and updates the import script for the MCUX SDK to correctly import RT1160 SDK files (using the same special import case as the RT1170)


galak (commented):
Do we really need the board files?

mcux/CMakeLists.txt
```diff
@@ -56,12 +62,12 @@ if (${MCUX_DEVICE} MATCHES "LPC|MIMXRT6")
   endif()
 endif()
 
-if (${MCUX_DEVICE} MATCHES "MIMXRT117")
+if (${MCUX_DEVICE} MATCHES "MIMXRT11")
```
galak, line 65: What case is this trying to cover?

danieldegrasse: The soc file for the RT11xx series (soc/arm/nxp_imx/rt/soc_rt11xx.c) makes use the the functions defined in these HAL drivers, whereas the soc file for the RT10xx series does not. This case pulls in those files for RT11xx devices to support the soc file.

As for why the SOC file itself calls those functions, it is based on the board initialization code within our MCUXpresso SDK.

danieldegrasse:
> 
> 
> Do we really need the board files?

@galak I believe the display_support file was added for future support, but the flexspi and dcd files are included by the CMakeLists.txt already. The dcd file is used for configuring the external memory, and flexspi one is used when the flexspi nor is selected as the boot device.

mmahadevan108:
> > @galak I believe the display_support file was added for future support, but the flexspi and dcd files are included by the CMakeLists.txt already. The dcd file is used for configuring the external memory, and flexspi one is used when the flexspi nor is selected as the boot device.
> 
> So the general rule of thumb has been not to have board specific code in the HALs like this. Some has slipped in. The dcd files should be moved into the zephyr tree. Take a look at `mm_feather` and `mm_swiftio`. We need to cleanup the mimxrt boards (been on my todo list).

We are moving to get the Zephyr HAL files from the SDK github repo. Moving these to Zephyr main repo would mean we have to maintain that copy rather than pulling it upates from the SDK github.

mmahadevan108:
> > We are moving to get the Zephyr HAL files from the SDK github repo. Moving these to Zephyr main repo would mean we have to maintain that copy rather than pulling it upates from the SDK github.
> 
> These are board files so non-NXP boards using NXP silicon will have to do that anyways. I don't see how the values that are DCD are likely to change that much that it should be an issue.
> 
> Additionally, as stated the Zephyr HAL policy is leaning towards not having board code in the HALs.

Makes sense to move these files to Zephyr. Will Zephyr Coding check make an exception for these files, else we may not be able to drop-in some of the files.

danieldegrasse:
The PR should be updated to remove all board specific files, and the accompanying Zephyr PR to support the RT1160 now has the required board files added.

PR #128, opened 2021-12-10
cmake: Fix include of sys cmake files for RT11xx

mmahadevan108:
System cmake files for RT11xx platforms are named differently


dleach02 (approved):
unfortunate, but I guess understandable since these are dual core platforms. How is this handled on the lpc55s69?

danieldegrasse (approved):

mmahadevan108:
> unfortunate, but I guess understandable since these are dual core platforms. How is this handled on the lpc55s69?

LPC55S69 has a similar naming convention i.e one per core.

PR #159, opened 2022-04-14
mcux: scripts: add lpc configuration scripts

danieldegrasse:
Add lpc configuration parsing scripts. These scripts are capable of
parsing a signal_configuration.xml file for an LPC or iMX RT 3 digit
part, and generating an output pin control header file as well as pin
control dtsi file containing pin mux options.


danieldegrasse:
@dleach02 The script is updated for v11 of the configuration tools now, so it should be good to merge.

danieldegrasse:
@mmahadevan108 Could you take a look at this? I'd like to merge it in when we can, since the pin control implementation for LPC should be stable at this point

PR #181, opened 2022-08-16
Update for the lpc55s36 IAP device driver

yvanderv:
Update the hal CMAKE file to reflect the device specific nature and location of the code.


dleach02 (approved):

mcux/mcux-sdk/devices/LPC55S36/drivers/driver_flashiap.cmake
```diff
@@ -1,6 +1,6 @@
 #Description: ROMAPI FLASHIAP Driver; user_visible: True
 include_guard(GLOBAL)
-message("driver_flashiap component is included.")
+message("driver_flashiap (lpc55s36) component is included.")
```
dleach02, line 4: So this is a change to a file provided by mcux_sdk... is this really necessary? Because if it is then you will need to note the change in the readme and let the SDK team know to make the appropriate changes in the SDK so we don't have to keep patching this in future updates.

yvanderv: No, it is not necessary per se, it makes the reading of the logs a lot clearer

yvanderv: Removed

mmahadevan108 (approved):

PR #182, opened 2022-08-17
Introduce initial support for NXP S32Z27 SoC

manuargue:
This PR introduces initial support for NXP S32 family, specifically for S32Z27 SoC from S32Z/E series.

The "drivers" folder contains device header files and bare metal peripheral drivers for NXP S32Z/E SoCs. Zephyr shim drivers are built on top of these imported drivers to adapt the S32 drivers APIs to Zephyr APIs. The "soc" directory contains SoC-specific default initialization (the configuration was generated using NXP S32 Design Studio for S32 Platform).


galak (changes requested):
Figure out how to implement the board code on the Zephyr side.  We should NOT have board code in HALs, it doesn't scale.

manuargue:
> Figure out how to implement the board code on the Zephyr side. We should NOT have board code in HALs, it doesn't scale.

Hi @galak, thanks for the feedback. Would it be acceptable to move it as it is to the board folder in zephyr repo? Some of the files, specially all the includes, are required for the SDK to build. Other sources, as the clock configuration, we need it to initialize clocks at boot-time since we are not providing a clock driver.

galak:
> > 
> 
> Hi @galak, thanks for the feedback. Would it be acceptable to move these board-related files as it is to the board folder in zephyr repo? Some of the files, specially all the includes, are required for the SDK to build. For other sources, such as the clock configuration, we need it to initialize clocks at boot-time since we are not providing a clock driver.

You'll need to 1. ensure license is Apache for the code to be in zephyr repo.  2. conforms to zephyr coding style guidelines at a minimum.  @dleach02 can probably help provide guidance on such things inside NXP.

What board code is required for the SDK to build?

clock config should ideally be handled by a clock control driver(s) in zephyr.  If not, there are numerous cases of doing soc init of clocks in `clock_init` function.

manuargue:
> What board code is required for the SDK to build?

All code in https://github.com/zephyrproject-rtos/hal_nxp/pull/182/commits/bc9c92eca4087cf531452acb57c8ac9d4b7df27d is auto-generated with a tool (NXP S32 Design Studio). This serve as default initialization for the peripherals and clocks, as well as to let the SDK drivers know which features are enabled/disabled and other set various defines. The SDK drivers are including most of the files in `generate/include`, so they must exist somewhere. That's the reason it was placed here instead of in zephyr main tree.

> clock config should ideally be handled by a clock control driver(s) in zephyr. If not, there are numerous cases of doing soc init of clocks in `clock_init` function.

Using the auto-generated clock configuration, we are initializing clocks at board init time (https://github.com/zephyrproject-rtos/zephyr/pull/49161/commits/3473c4d9b0853ed5f760b33adb8c41117bfa60e4#diff-40b1e148b7756fd3abea2a33d9d2f6408b481907b97ba03dfcb9e05afa4a775dR23). Currently we don't have a clock driver in Zephyr and we rely on this static configuration.

For the rest of the peripheral drivers, e.g. UART, the shim drivers are using the generated structs (`generate/src`) to initialize the peripheral with the default configuration. For now, the ability to configure them from DT is minimal.

dleach02:
@galak, please revisit your change request to determine if it was satisfactorily addressed

s32/drivers/s32ze/Dio/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+# Copyright 2022 NXP
+# SPDX-License-Identifier: BSD-3-Clause
+
+zephyr_include_directories(include)
```
galak, line 4: Is this really needed?  (same question for any CMakeLists.txt) file.  I'd assume the header in here is local.

manuargue: these are the driver headers included in the shim drivers, e.g. https://github.com/zephyrproject-rtos/zephyr/pull/49161/files#diff-76940438b7185f8b5531917a8b22b2ae58b0f37ecbbf01c08ab31acc81c1babdR12-R13

manuargue:
> I'm guessing `s32/scripts/gen_s32ct_defines.py` isn't part of the generated files, so would suggest pull it into its own commit, so that there is one commit that is just adding the generated files. (Guess that goes for Cmakefiles, etc..). Would like to see one commit which is just that files that come from "NXP S32 Design Studio", so its clear what is coming from that tool and what was added as integration.

I have split the changes in multiple commits as requested.

galak (changes requested):
So I'm guessing all the CMakeLists.txt files are not part of what the SDK/tool generate, if that is true, they should be in their own commits.

So it should be:
* commit(s) that add the files that the SDK/tool generate
* commit that adds the CMakeLists.txt files
* commit(s) that modify any of the SDK files

s32/CMakeLists.txt
```diff
@@ -22,3 +22,32 @@ zephyr_compile_definitions(${S32_DEVICE})
 
 add_subdirectory(drivers/${DRIVERS_BASE})
 add_subdirectory(soc/${SOC_BASE})
+
+# Some shim drivers use Kconfig string options to glue with driver configuration.
```
galak, line 26: How/where is s32ct_define.h used?

s32/drivers/s32ze/Dio/CMakeLists.txt
galak: Can we simplify this to just be at the `drivers/s32ze` level so we can replace:

```
-I/Users/galak/git/modules/hal/nxp/s32/drivers/s32ze/BaseNXP/header 
-I/Users/galak/git/modules/hal/nxp/s32/drivers/s32ze/BaseNXP/include
-I/Users/galak/git/modules/hal/nxp/s32/drivers/s32ze/Dio/include
-I/Users/galak/git/modules/hal/nxp/s32/drivers/s32ze/Mcu/include
-I/Users/galak/git/modules/hal/nxp/s32/drivers/s32ze/Port/include 
-I/Users/galak/git/modules/hal/nxp/s32/drivers/s32ze/Rte/include 
-I/Users/galak/git/modules/hal/nxp/s32/drivers/s32ze/Uart/include 
```

with
```
-I/Users/galak/git/modules/hal/nxp/s32/drivers/s32ze
```

and tweak the `#include`s to be `#include <Port/include/Siul2_Port_Ip.h>`

s32/CMakeLists.txt
galak: Glancing at how this is getting used, we still seem to have "BOARD" bits in `s32/soc/s32z27/src/Linflexd_Uart_Ip_Sa_BOARD_InitPeripherals_PBcfg.c`.   Can we not just have instances of  `Linflexd_Uart_Ip_UserConfigType` setup in the shim driver.  What happens when someone wants needs a second or third UART?

s32/drivers/s32ze/Dio/CMakeLists.txt
manuargue: I agree this is a bit verbose. Although these headers are not only included by the shim drivers but also by the SDK baremetal drivers as well, so following suggested change implies tweaking all SDK driver includes as well. I prefer to avoid modifications to the SDK drivers unless necessary, to reduce maintenance effort when updating the SDK version or adding new baremetal drivers.

manuargue:
> So I'm guessing all the CMakeLists.txt files are not part of what the SDK/tool generate, if that is true, they should be in their own commits.

@galak I have further separated the changes into multiple commits so it's easier to know the origin of the sources. All the CMakeLists.txt files are added as part of the integration so they I have kept them in a single commit.

s32/CMakeLists.txt
manuargue: > How/where is s32ct_define.h used?

I have pushed the following changes:
- renamed the generated header from `s32ct_define.h` to `s32ct_conf.h`, which I consider more appropriate
- removed the global include through imacros of (now) `s32ct_conf.h` and include this header in each shim driver source as needed (e.g. https://github.com/zephyrproject-rtos/zephyr/pull/49161/commits/245876f67ed631f772c4040365aef54524815e9b#diff-210804e4917b7166b3e91bff5b45853537b6468078848fa45bdcc8b6bf5d4499R10)

At the moment this auto-generated header is used only by the serial driver, to build a macro with the name of the default configuration for Uart (https://github.com/zephyrproject-rtos/zephyr/pull/49161/commits/245876f67ed631f772c4040365aef54524815e9b#diff-210804e4917b7166b3e91bff5b45853537b6468078848fa45bdcc8b6bf5d4499R332). The use is not limited to this case, but other peripheral drivers (that we plan on open a PR after this one) make use of it as well.

manuargue: > Glancing at how this is getting used, we still seem to have "BOARD" bits in s32/soc/s32z27/src/Linflexd_Uart_Ip_Sa_BOARD_InitPeripherals_PBcfg.c

The "BOARD_InitPeripherals" is just a default name set by the configuration tool. The code generated doesn't depend on the Zephyr board being built. So future boards based on the same SoC can still use the generated code.

> Can we not just have instances of Linflexd_Uart_Ip_UserConfigType setup in the shim driver. What happens when someone wants needs a second or third UART?

I have reworked the integration between the SDK drivers, their default configuration and Zephyr shim drivers, to align better with the approach followed by other HAL's and remove the need for the user to use the configurator tools to enable other instances than the ones available in the default configuration.

The SDK driver default configuration serves now as template to populate at run-time (during device initialization) the init configuration for the peripheral instances enabled through devicetree. For example, https://github.com/zephyrproject-rtos/zephyr/pull/49161/commits/178eaf526754841d72a5d7e8dfbbaa0d1d22d4fd#diff-210804e4917b7166b3e91bff5b45853537b6468078848fa45bdcc8b6bf5d4499R296-R300. In the case of UART, the instance configuration must be stored in the device data struct because the SDK driver will keep a reference to it. This may not be required for other peripheral drivers, and the instance configuration could live in the stack of the device init function to save memory.

> How/where is s32ct_define.h used?

Due to the rework, this auto-generated header and the script generating it, are no longer necessary.

galak (approved):

s32/CMakeLists.txt
galak: I still dont see how or why we need `Linflexd_Uart_Ip_xHwConfigPB_0_BOARD_INITPERIPHERALS` and thus `Linflexd_Uart_Ip_Sa_BOARD_InitPeripherals_PBcfg.c`.  It seems like how `Linflexd_Uart_Ip_xHwConfigPB_0_BOARD_INITPERIPHERALS` is defined will ONLY work for UART0, since it has `&Linflexd_Uart_Ip_apStateStructure[0]`.  Doesn't that need to change based on the UART being used?

galak (changes requested):
Still a question about use of Linflexd_Uart_Ip_xHwConfigPB_0_BOARD_INITPERIPHERALS

s32/CMakeLists.txt
manuargue: `Linflexd_Uart_Ip_Sa_BOARD_InitPeripherals_PBcfg.c` is a file generated by the configuration tool and contains initialization code and state structures for the UART peripheral. `Linflexd_Uart_Ip_xHwConfigPB_0_BOARD_INITPERIPHERALS` is the struct used to initialize UART instance 0, but we are using it as a "template" for the initial configuration of all UART instances. The behavior is _somehow analogous_ to the `*_GetDefaultConfig()` functions used by MCUX to populate the peripheral initialization struct. The UART shim driver uses this template struct with default configuration to create its own private copy (per device instance), modifying the struct fields as needed. The only modification to the default configuration being done at the moment, is to update the reference to the UART state struct (`&Linflexd_Uart_Ip_apStateStructure[0]`) based on the device instance initialized. We plan in the next PR's to introduce the ability to modify more fields through devicetree properties. You can see current implementation in https://github.com/zephyrproject-rtos/zephyr/blob/178eaf526754841d72a5d7e8dfbbaa0d1d22d4fd/drivers/serial/uart_s32_linflexd.c#L296-L300.

The tool-generated code for UART is rather simple, but for other peripherals that we plan to support, the generated code is more complex (e.g. peripherals that interact with each other) and there are a few configuration structs per instance. This is one of the reasons we have kept `Linflexd_Uart_Ip_Sa_BOARD_InitPeripherals_PBcfg.c` as it is generated by the tool for the time being. Another reason is that is going to be easier do updates of the tool-generated code. Nevertheless, after the feedback on this PR we will continue exploring a better approach to perhaps depend less on the configuration tool and overcome any shortcoming of current implementation.

manuargue:
> Still a question about use of Linflexd_Uart_Ip_xHwConfigPB_0_BOARD_INITPERIPHERALS

Replied in https://github.com/zephyrproject-rtos/hal_nxp/pull/182#discussion_r975980637

s32/CMakeLists.txt
galak: W/o seeing what generated code for the other peripherals looks like its hard to comment on them.  I'd say this is different from `*_GetDefaultConfig()` since that isn't generated and static.  I'd suggest it makes more sense to just setup `Linflexd_Uart_Ip_xHwConfigPB_0_BOARD_INITPERIPHERALS` in the shim driver and not use the generated code as a template for the UART.  We can cross the bridge as we come to it for other peripherals as those are supported.

galak:
> > Still a question about use of Linflexd_Uart_Ip_xHwConfigPB_0_BOARD_INITPERIPHERALS
> 
> Replied in [#182 (comment)](https://github.com/zephyrproject-rtos/hal_nxp/pull/182#discussion_r975980637)

Looking for a reply to https://github.com/zephyrproject-rtos/hal_nxp/pull/182#discussion_r976874141

s32/CMakeLists.txt
galak: I think thats it at this point.

manuargue:
@galak I've pushed the following updates:

- Remove tool generated `Linflexd_Uart_Ip_Sa_BOARD_InitPeripherals_PBcfg.c` and move default UART configuration to shim driver (https://github.com/zephyrproject-rtos/zephyr/blob/ad95d0451cbd3c7fd52a42f85c93c9acc3ca5e2f/drivers/serial/uart_s32_linflexd.c#L296-L309).
- Renamed default name for tool generated files `*BOARD_InitPeripherals*` to  `*Init*`, just for the sake of not creating confusion that these files are related to the boards.

galak:
> * Renamed default name for tool generated files `*BOARD_InitPeripherals*` to  `*Init*`, just for the sake of not creating confusion that these files are related to the boards.

Is there any reason to have `Linflexd_Uart_Ip_Sa_Init_PBcfg.h`?

manuargue:
> > * Renamed default name for tool generated files `*BOARD_InitPeripherals*` to  `*Init*`, just for the sake of not creating confusion that these files are related to the boards.
> 
> Is there any reason to have `Linflexd_Uart_Ip_Sa_Init_PBcfg.h`?

It's included in https://github.com/zephyrproject-rtos/hal_nxp/blob/269f66c7edacd9b2c42dc8cb56ffd7d51af95785/s32/soc/s32z27/include/Linflexd_Uart_Ip_Cfg.h#L32
Of course, include can be removed. But I would like to avoid modifying manually to many tool-generated files..

galak:
> Of course, include can be removed. But I would like to avoid modifying manually to many tool-generated files..

can we have the file just be empty than?

manuargue:
> > Of course, include can be removed. But I would like to avoid modifying manually to many tool-generated files..
> 
> can we have the file just be empty than? With some comment about how its normally generated, but not used by Zephyr integration.

In order to make it an empty file I need to manually modify 3 other files, including a driver header `Uart\include\Linflexd_Uart_Ip.h`. Plus, the SDK uses some macros to cross-check the file versions. So the minimal, `Linflexd_Uart_Ip_Sa_Init_PBcfg.h` would be:
```c
// all the version macros must be kept
#define LINFLEXD_UART_IP_SA_PBCFG_*_VERSION              *
// macro required, but can be modified to be empty
#define LINFLEXD_UART_IP_SA_CONFIG_INIT_PB
```
Honestly there's really not much difference with the current generated code. Btw, I correct myself, removing the include for this header has the same effect as making it empty.

galak:
@manuargue I'd say go with the minimal version of the file.

manuargue:
> @manuargue I'd say go with the minimal version of the file.

Ok. I've pushed it @galak

galak:
@manuargue can you update the Zephyr side PR to get it to re-run CI.  I don't think CI will re-run when the HAL side changes.  Just to make sure things build.  Once that is done, I'll approve the HAL side.

galak (approved):

dleach02 (approved):

danieldegrasse (approved):

mmahadevan108 (approved):

PR #186, opened 2022-08-29
Upstream Cortex-A Core ARM64 support

JiafeiPan:
This patchset is sync harpoon v2.1.0 release from mcux-sdk to hal_nxp.


mcux/mcux-sdk/drivers/enet/fsl_enet.c
```diff
@@ -793,7 +794,8 @@ static void ENET_SetTxBufferDescriptors(enet_handle_t *handle,
     uint8_t ringNum;
     uint16_t count;
     uint32_t txBuffSizeAlign;
-    uint8_t *txBuffer                   = NULL;
+    uintptr_t txBuffer                  = 0;
```
dleach02, line 797: what is wrong with NULL?

JiafeiPan: This update change txBuffer from an pointer to "uintptr_t" type which is uint64 for ARM64 and uint32_t for ARM32, it is to fx the size of 'buffer' in RX and TX buffer descriptors, the buffer pointer has fixed size of 4 Bytes in buffer descriptors, but the current type of 'buffer' is 'uint8_t *' , it isn't compatible with 64-bit platforms.

hakehuang (commented):
harpoon is our internal project name, better to change to official name

JiafeiPan:
> harpoon is our internal project name, better to change to official name

Updated, thanks.

mmahadevan108 (changes requested):
These changes should go into https://github.com/NXPmicro/mcux-sdk

It is hard to maintain local changes here especially when we have to update the SDK.

JiafeiPan:
> These changes should go into https://github.com/NXPmicro/mcux-sdk
> 
> It is hard to maintain local changes here especially when we have to update the SDK.

All these patches in this PR have already merged into mcux-sdk main branch.

hakehuang:
@JiafeiPan CI testing OK, no issue found.

hakehuang (approved):

mmahadevan108 (approved):

PR #196, opened 2022-10-06
mcux-sdk: s/ARRAY_SIZE/FSL_ARRAY_SIZE

gmarull:
ARRAY_SIZE is already defined by Zephyr, so it can't be re-used in public headers such as fsl_common.h (used by some Zephyr code). Rename the macro to FSL_ARRAY_SIZE instead.

Automated using:

```
perl -i -pe 's/([^_])ARRAY_SIZE\((.*)\)/\1FSL_ARRAY_SIZE(\2)/g' **/*.c/h
```

Signed-off-by: Gerard Marull-Paretas <gerard.marull@nordicsemi.no>

### DIFF
--- a/mcux/mcux-sdk/boards/evkmimxrt595/board.c
+++ b/mcux/mcux-sdk/boards/evkmimxrt595/board.c
@@ -236,7 +236,7 @@ status_t BOARD_InitPsRam(void)
     FLEXSPI_SetFlashConfig(BOARD_FLEXSPI_PSRAM, &deviceconfig, kFLEXSPI_PortA1);
 
     /* Update LUT table. */
-    FLEXSPI_UpdateLUT(BOARD_FLEXSPI_PSRAM, 0, customLUT, ARRAY_SIZE(customLUT));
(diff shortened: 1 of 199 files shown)


dleach02:
@gmarull, while I appreciate what you are trying to do here, it is something we don't want to pull into our HAL due to how our HAL is sourced and constructed. This type of change would need to be contributed to the source mcu SDK for those maintainers to consider.

We have been avoiding this problem in our Zephyr enablement by ensuring the order of include files. We also have protection around our definition to avoid the redefinition problem that occurs if this is included before the Zephyr version.

gmarull:
Let’s not close this until we have the discussion in arch wg. If Zephyr doesn’t want to namespace its own ARRAY_SIZE I’m afraid 3rd party code will need to be patched to avoid namespace pollution. Protecting against redefinition is bad, since final implementation depends on include order, and we should not introduce yet another bad practice with includes because of this.

dleach02:
@gmarull, I'm closing this again since you provided a less intrusive mechanism for dealing with the Zephyr problem.

PR #197, opened 2022-10-18
mcux-sdk: use Zephyr's MIN/MAX/ARRAY_SIZE

gmarull:
When HAL is used in the context of Zephyr, use Zephyr's MIN/MAX/ARRAY_SIZE to avoid namespace collisions.

Also moved KB definition to the only source file using it (was also causing collisions in Zephyr).

Signed-off-by: Gerard Marull-Paretas <gerard.marull@nordicsemi.no>

### DIFF
--- a/mcux/mcux-sdk/CMSIS/Core_A/Include/mmu_armv8a.h
+++ b/mcux/mcux-sdk/CMSIS/Core_A/Include/mmu_armv8a.h
@@ -41,10 +41,6 @@
  extern "C" {
 #endif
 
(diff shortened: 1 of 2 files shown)


gmarull:
Alternative to https://github.com/zephyrproject-rtos/hal_nxp/pull/196

stephanosio (approved):

de-nordic (approved):
Great. Love the idea of removing common defines out of interface headers.

dleach02 (approved):

PR #215, opened 2022-12-22
mcux: hal_nxp: only relocate flexSPI driver when CONFIG_FLASH=y

danieldegrasse:
Only relocate flexSPI driver when CONFIG_FLASH=y.
CONFIG_FLASH_MCUX_FLEXSPI_XIP previously was dependent on CONFIG_FLASH, but the scope of this Kconfig has changed. Due to this, the flexSPI driver now must have two checks, as it should not relocated when the driver is being used for a memory controller that does not expose the flash driver interface


danieldegrasse:
@EmilioCBen FYI

PR #224, opened 2023-01-12
mcux: include device_system for the LPC55XXX

EmilioCBen:
Added the Kconfig check for LPC55 Family
when trying to include device_system.
This will allow the LPC55XXX soc.c
to utilize SystemCoreClock

Signed-off-by: Emilio Benavente <emilio.benavente@nxp.com>


decsny:
I am a bit confused what this variable is being used for, I see that you want to write to it in the soc file in zephyr, but if we currently aren't using it and it is just a uint32 global variable, why can you not just declare it in the soc file. That seems a lot simpler than revising the HAL and building an entire new file just for the one int variable

EmilioCBen:
> I am a bit confused what this variable is being used for, I see that you want to write to it in the soc file in zephyr, but if we currently aren't using it and it is just a uint32 global variable, why can you not just declare it in the soc file? That seems a lot simpler than revising the HAL and building an entire new file just for the one int variable

The SystemCoreClock variable, if that's what you are referring to, is being used to handle interrupts in

> I am a bit confused what this variable is being used for, I see that you want to write to it in the soc file in zephyr, but if we currently aren't using it and it is just a uint32 global variable, why can you not just declare it in the soc file? That seems a lot simpler than revising the HAL and building an entire new file just for the one int variable

The SystemCoreClock variable is being used to handle the delay on the board, and the soc file is applying potentially the correct value pre board, via the SOC family relation.

decsny:
I guess one reasoning is that this file will probably be built since CONFIG_PLATFORM_SPECIFIC_INIT will be set most likely from Daniel's changes and declaring this variable from the soc.c will probably cause a conflict, and to be honest I think this file should always be included for every SOC anyways

But I still don't understand what we are talking about when we are saying this variable generates delays in the board, the Zephyr kernel uses its own data for that. If the HAL is using it then I am confused why there was no problem before now when it wasn't being included. I would like to know what exactly we mean here just for future reference. But it doesn't matter as far as merging this because like I said I think the file should be included anyways, although the way we are doing it here looks very disorganized but that seems to be the theme for the hal_nxp.cmake file

EmilioCBen:
> I guess one reasoning is that this file will probably be built since CONFIG_PLATFORM_SPECIFIC_INIT will be set most likely from Daniel's changes and declaring this variable from the soc.c will probably cause a conflict, and to be honest I think this file should always be included for every SOC anyways
> 
> But I still don't understand what we are talking about when we are saying this variable generates delays in the board, since the Zephyr kernel uses its own data/variables for that. If the HAL is using it then I am confused why there was no problem before now, when it wasn't being included. I would like to know what exactly we mean here just for future reference. But it doesn't matter as far as merging this because like I said I think the file should be included anyways, although the way we are doing it here looks very disorganized but that seems to be the theme for the hal_nxp.cmake file


@decsny 
Here are a few examples where the MCUX-SDK use the SystemCoreClock variable to calculate a Delay time.
SDMMC -> https://github.com/nxp-mcuxpresso/mcux-sdk-examples/blob/95e0e9b2aa9d27cec8f12842ee1146792d5b8311/lpcxpresso55s69/azure_rtos_examples/filex_sdcard/cm33_core0/sdmmc_config.c#L36

AzureRTOS -> https://github.com/nxp-mcuxpresso/mcux-sdk-examples/blob/95e0e9b2aa9d27cec8f12842ee1146792d5b8311/lpcxpresso55s69/azure_rtos_examples/uart_example/cm33_core0/board_setup.c#L14

Led_Blinky -> https://github.com/nxp-mcuxpresso/mcux-sdk-examples/blob/95e0e9b2aa9d27cec8f12842ee1146792d5b8311/lpcxpresso55s69/demo_apps/led_blinky/cm33_core0/led_blinky.c#L64

decsny:
> > I guess one reasoning is that this file will probably be built since CONFIG_PLATFORM_SPECIFIC_INIT will be set most likely from Daniel's changes and declaring this variable from the soc.c will probably cause a conflict, and to be honest I think this file should always be included for every SOC anyways
> > But I still don't understand what we are talking about when we are saying this variable generates delays in the board, since the Zephyr kernel uses its own data/variables for that. If the HAL is using it then I am confused why there was no problem before now, when it wasn't being included. I would like to know what exactly we mean here just for future reference. But it doesn't matter as far as merging this because like I said I think the file should be included anyways, although the way we are doing it here looks very disorganized but that seems to be the theme for the hal_nxp.cmake file
> 
> @decsny Here are a few examples where the MCUX-SDK use the SystemCoreClock variable to calculate a Delay time. SDMMC -> https://github.com/nxp-mcuxpresso/mcux-sdk-examples/blob/95e0e9b2aa9d27cec8f12842ee1146792d5b8311/lpcxpresso55s69/azure_rtos_examples/filex_sdcard/cm33_core0/sdmmc_config.c#L36
> 
> AzureRTOS -> https://github.com/nxp-mcuxpresso/mcux-sdk-examples/blob/95e0e9b2aa9d27cec8f12842ee1146792d5b8311/lpcxpresso55s69/azure_rtos_examples/uart_example/cm33_core0/board_setup.c#L14
> 
> Led_Blinky -> https://github.com/nxp-mcuxpresso/mcux-sdk-examples/blob/95e0e9b2aa9d27cec8f12842ee1146792d5b8311/lpcxpresso55s69/demo_apps/led_blinky/cm33_core0/led_blinky.c#L64

Okay.... I am not saying you are wrong but I was just wondering why this particular variable is so important from a zephyr HAL standpoint... but I am not sure why some azure RTOS examples are useful to zephyr

The thing is I think the file should be included anyways, but I just was wondering why this variable is important, which it seems like it is, since you are making another PR and adding to the Cmake code just to declare it

dleach02:
Why wouldn't the CONFIG_PLATFORM_SPECIFIC_INIT be enough for LPC55XXXX? 

It seems like you are trying to force the platforms down the path and for some unknown reason the CONFIG_PLATFORM_SPECIFIC_INIT is not set for the SOC/board, which would have also forced things down this path? I guess I would say the same about the existing check for CONFIG_SOC_SERIES_IMX_RT6XX in the 'if' statement.

dleach02 (approved):
we discussed my questions in a different forum. There is a larger activity that will clean this up later so for expediency of the PR will approve this.

mmahadevan108 (approved):

danieldegrasse (approved):

PR #231, opened 2023-04-11
mcux_sdk: include vref_1 if using LPC55S36

decsny:
Include vref_1 if using lpc55s36


danieldegrasse:
why not include the VREF1 driver based on the Kconfig that enables this driver in Zephyr? Similar to how we enable drivers for other peripherals: https://github.com/zephyrproject-rtos/hal_nxp/blob/master/mcux/hal_nxp.cmake#L98

danieldegrasse (approved):
Declan confirmed with me that the VREF can be a separate driver, but that this will be future work. Approving this as it works for a short term solution

decsny:
> why not include the VREF1 driver based on the Kconfig that enables this driver in Zephyr? Similar to how we enable drivers for other peripherals: https://github.com/zephyrproject-rtos/hal_nxp/blob/master/mcux/hal_nxp.cmake#L98

There is no Kconfig for this driver in Zephyr (yet). When I add a driver for VREF it will have a Kconfig but not be included here still because it will be a native zephyr driver. I will revert this commit when I publish that branch as a PR.

dleach02 (approved):

PR #236, opened 2023-04-25
mcux: cmake: include pixel pipeline (PXP)

dariis:
Enable PXP HAL driver as Zephyr DMA driver.


mmahadevan108:
@dariis , thank you for your contribution. Is there a Zephyr PR that your working on to go along with this change in the HAL

dariis:
Unfortunately not. I implemented a dummy pxp dma driver with only an init and the real implementation the hacky way in the eLCDIF driver.

mcux/hal_nxp.cmake
```diff
@@ -129,6 +129,7 @@ include_driver_ifdef(CONFIG_COUNTER_MCUX_PIT		pit		driver_pit)
 include_driver_ifdef(CONFIG_COUNTER_MCUX_QTMR		qtmr_1		driver_qtmr_1)
 include_driver_ifdef(CONFIG_DISPLAY_MCUX_ELCDIF		elcdif		driver_elcdif)
 include_driver_ifdef(CONFIG_DISPLAY_MCUX_DCNANO_LCDIF	lcdif		driver_lcdif)
+include_driver_ifdef(CONFIG_DMA_MCUX_PXP		pxp		driver_pxp)
```
danieldegrasse, line 132: Since this driver isn't currently integrated into Zephyr as DMA driver could we use `CONFIG_MCUX_PXP` as the Kconfig name? This keeps the Kconfig generic enough for future uses.

dariis: IMHO that has to be decided by the maintainers if it need to match to a Zephyr subsystem. I can change it as you wish.

mmahadevan108: @dariis could you please help make the change.

mmahadevan108 (approved):

PR #237, opened 2023-04-26
mcux: mcux-sdk: fix calculation of Flexcomm USART BRG value in 32K mode

danieldegrasse:
Fix calculation of the USART BRG value when using the flexcomm in 32KHz clock mode. This value should be one less, or the resulting baud rate will be half what the user requests. This issue is already fixed in the USART_Enable32kMode() function, but still existed in the USART_Init() function.


mmahadevan108:
Can you please send a PR to the SDK Github

mmahadevan108:
Also is there a Zephyr PR for this?

danieldegrasse:
Closing this PR, as we are not currently planning to enable the low power mode that would utilize 32K mode, and the fix for this issue is already present within the SDK (so should be brought into the HAL on the next SDK update)

PR #254, opened 2023-07-10
Add support for FLexCAN on S32K344

congnguyenhuu:
Add FlexCAN instance base addresses and module features to support using FlexCAN MCUX driver.

### DIFF
--- a/s32/CMakeLists.txt
+++ b/s32/CMakeLists.txt
@@ -24,6 +24,9 @@ if(CONFIG_HAS_MCUX)
   #
   # Glue code must be added for each device supported to build with MCUX. This
   # can be found in hal_nxp/s32/mcux/devices:
+  # - s32/mcux/devices/${MCUX_DEVICE}/${MCUX_DEVICE}_device.h
+  #   redefine RTD module's peripheral access layers and its register masks to be
+  #   compatible with MCUX drivers for incompatible peripherals
   # - s32/mcux/devices/${MCUX_DEVICE}/${MCUX_DEVICE}_features.h
   #   define SoC module's features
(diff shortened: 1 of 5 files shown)


mmahadevan108 (approved):

congnguyenhuu:
Hi @mmahadevan108 @dleach02, If  there is not any comment. Could we merge this PR? so that it can be got in Zephyr PR. Reveiwers are looking forward to that. Thanks !

congnguyenhuu:
> @congnguyenhuu please add to the docs in https://github.com/zephyrproject-rtos/hal_nxp/blob/master/s32/CMakeLists.txt#L19-L42, the new file being introduced `s32/mcux/devices/S32K344/S32K344_device.h`. Ideally we should have this documentation somewhere else in a README, but for now that's the place.

I updated the docs for it

s32/CMakeLists.txt
```diff
@@ -24,6 +24,9 @@ if(CONFIG_HAS_MCUX)
   #
   # Glue code must be added for each device supported to build with MCUX. This
   # can be found in hal_nxp/s32/mcux/devices:
+  # - s32/mcux/devices/${MCUX_DEVICE}/${MCUX_DEVICE}_device.h
+  #   redefine RTD module's peripheral access layers and its register masks to be
+  #   compatible with MCUX drivers for incompatible peripherals (FlexCAN)
```
manuargue, line 29: just a minor comment, please remove (FlexCAN) since this is intended to be a generic description. We may need to do the same for other peripherals as well.

congnguyenhuu: Okay, I updated. I thought that need to note specific peripherals because it is not all peripherals need to be done.

manuargue (approved):

dleach02 (approved):

PR #259, opened 2023-07-31
mcux: scripts: update soc header script to handle cfg reg type correctly

danieldegrasse:
Update SOC header script for IMX parts to handle configuration register
type correctly. RT11xx SOCs have multiple types of configuration
registers for pins, and this needs to be encoded in the devicetree for
the MUX options. Fix the header generation script to correctly evaluate
the config type based on which bitfields are in the register, instead of
based on the pin name (as this pattern is not valid for all GPIO_DISP
pins)

Also, update RT1170 and RT1160 header options to have correct pin type
information.

Note that due to changes in source data, the following changes have also
been made to the headers:
- the names of SNVS pins have been updated to align with the RM
- XBAR_INOUT mux options now set the associated GPR bit to select the
  XBAR pin as an output, while XBAR_IN options leave the GPR bit clear.


dts/nxp/nxp_imx/rt/mimxrt1176dvmaa-pinctrl.dtsi
```diff
@@ -6277,104 +6272,116 @@
...
 	};
-	/omit-if-no-ref/ iomuxc_snvs_pmic_stby_req_gpio13_io02: IOMUXC_SNVS_PMIC_STBY_REQ_GPIO13_IO02 {
-		pinmux = <0x40c94008 5 0x0 0 0x0>;
+	/omit-if-no-ref/ iomuxc_snvs_pmic_stby_req_dig_gpio13_io02: IOMUXC_SNVS_PMIC_STBY_REQ_DIG_GPIO13_IO02 {
+		pinmux = <0x40c94008 5 0x0 0 0x40c94048>;
 		pin-snvs;
 	};
-	/omit-if-no-ref/ iomuxc_snvs_pmic_stby_req_pgmc_pmic_vstby_req: IOMUXC_SNVS_PMIC_STBY_REQ_PGMC_PMIC_VSTBY_REQ {
-		pinmux = <0x40c94008 0 0x0 0 0x0>;
+	/omit-if-no-ref/ iomuxc_snvs_pmic_stby_req_dig_pgmc_pmic_vstby_req: IOMUXC_SNVS_PMIC_STBY_REQ_DIG_PGMC_PMIC_VSTBY_REQ {
```
danieldegrasse, line 6371: Slightly concerned about the pin name changes here- this aligns with the source configuration data used to create these headers (and with the reference manual), but changing the naming here will break builds for customers using these pins downstream

mmahadevan108: We should document in the Zephyr release notes.

PR #261, opened 2023-08-08
mcux: add tempmon driver to cmake

blemouzy:
Add tempmon driver to MCUX cmake directive, to enable inclusion of MCUX TEMPMON HAL driver.


blemouzy:
> @blemouzy since you are not using this hal driver in your linked zephyr PR, can we close this for now?

Yes, we can.

blemouzy:
Zephyr driver will not use HAL code.

PR #267, opened 2023-08-15
license: change location of the NXP HAL license file

yvanderv:
The NXP HAL is licensed under the terms of BSD-3. For clarity, the license file is moved to the root of the NXP HAL repository


mmahadevan108:
@yvanderv, do we need to move this file every time we update the SDK?

yvanderv:
> @yvanderv, do we need to move this file every time we update the SDK?

well, we will have to be careful when we do the HAL upgrades.

hakehuang (approved):

decsny:
> well, we will have to be careful when we do the HAL upgrades.

should we update the README then

mmahadevan108 (approved):

PR #283, opened 2023-09-21
Fix mpid redefined

JiafeiPan:
Sync MPIDR_AFFLVL_MASK with Zephyr to avoid building warning for redefinition.

### DIFF
--- a/mcux/mcux-sdk/CMSIS/Core_AArch64/Include/core_common.h
+++ b/mcux/mcux-sdk/CMSIS/Core_AArch64/Include/core_common.h
@@ -69,9 +69,7 @@
 #define GET_EL(_mode)		(((_mode) >> MODE_EL_SHIFT) & MODE_EL_MASK)
 
 /* MPIDR */
-#ifndef MPIDR_AFFLVL_MASK
 #define MPIDR_AFFLVL_MASK	(0xfful)
-#endif
 #define MPIDR_AFF0_SHIFT	(0)
 #define MPIDR_AFF1_SHIFT	(8)
 #define MPIDR_AFF2_SHIFT	(16)
--- a/mcux/mcux-sdk/CMSIS/Core_AArch64/Include/core_common.h
+++ b/mcux/mcux-sdk/CMSIS/Core_AArch64/Include/core_common.h
@@ -69,7 +69,7 @@
 #define GET_EL(_mode)		(((_mode) >> MODE_EL_SHIFT) & MODE_EL_MASK)
 
 /* MPIDR */
-#define MPIDR_AFFLVL_MASK	(0xfful)
+#define MPIDR_AFFLVL_MASK	(0xffULL)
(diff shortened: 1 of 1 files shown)


mcux/mcux-sdk/CMSIS/Core_AArch64/Include/core_common.h
```diff
@@ -69,9 +69,7 @@
 #define GET_EL(_mode)		(((_mode) >> MODE_EL_SHIFT) & MODE_EL_MASK)
 
 /* MPIDR */
-#ifndef MPIDR_AFFLVL_MASK
-#define MPIDR_AFFLVL_MASK	(0xfful)
-#endif
+#define MPIDR_AFFLVL_MASK	(0xffULL)
```
fabiobaltieri, line 72: Ok so you are matching the internal define so that the compiler does not complain about the redefinition right? I guess it's ok, certainly avoids the the include order problem but would it be an option to include the upstream header instead and not define these in the first place if that's an option?

cc @gmarull for a second opinion

gmarull: if the mask needs to be ULL then this patch is an actual bug fix, and if it luckily matches Zephyr define we can live with that. We are living with vendor HALs anyway :hankey:

fabiobaltieri: Alrighty, then I guess that's what I should have done yesterday instead of ifdef'ing it. Okay, live and learn.

fabiobaltieri:
Looks good, thanks for following up it. I have no permissions to get an actual green button to press in this repo, but if I did I would press it. :-)

mcux/mcux-sdk/CMSIS/Core_AArch64/Include/core_common.h
JiafeiPan: Yes, indeed, it should be ULL, and we have no building warning with this update.

JiafeiPan:
> Looks good, thanks for following up it. I have no permissions to get an actual green button to press in this repo, but if I did I would press it. :-)

Thanks.

mcux/mcux-sdk/CMSIS/Core_AArch64/Include/core_common.h
mmahadevan108: @JiafeiPan. Is there a way to avoid this duplication. Maybe move this code to the Zephyr CMSIS module?

JiafeiPan: @mmahadevan108, in the long term, we need to upstream CMSIS and then migrate to Zephyr's CMSIS, currently we need still to use hal's CMSIS before that,  and not able to only move part of code to another CMSIS.

mmahadevan108 (approved):

PR #304, opened 2023-11-23
drivers: Introduce revision 2 of the edma driver

LaurentiuM1234:
This is needed because currently there's no EDMA API that can be used for all eDMA IPs found
on i.MX93, i.MX8QM, and i.MX8QXP.

This API also allows working with different eDMA variants
found on the same SoC (i.MX93 with eDMA3 and eDMA4 which are different)

Depends on: https://github.com/nxp-mcuxpresso/mcux-sdk/pull/152

### DIFF
--- a/mcux/hal_nxp.cmake
+++ b/mcux/hal_nxp.cmake
@@ -215,6 +215,7 @@ include_driver_ifdef(CONFIG_DMA_MCUX_SMARTDMA		smartdma	driver_lpc_smartdma)
 include_driver_ifdef(CONFIG_DAC_MCUX_LPDAC			dac_1		driver_dac_1)
 include_driver_ifdef(CONFIG_NXP_IRQSTEER			irqsteer	driver_irqsteer)
 include_driver_ifdef(CONFIG_AUDIO_DMIC_MCUX		dmic		driver_dmic)
+include_driver_ifdef(CONFIG_DMA_NXP_EDMA	edma		driver_edma_rev2)
 
 if ((${MCUX_DEVICE} MATCHES "MIMXRT1[0-9][0-9][0-9]") AND (NOT (CONFIG_SOC_MIMXRT1166_CM4 OR CONFIG_SOC_MIMXRT1176_CM4)))
   include_driver_ifdef(CONFIG_HAS_MCUX_CACHE		cache/armv7-m7	driver_cache_armv7_m7)
--- a/mcux/mcux-sdk/devices/MIMX9352/drivers/fsl_edma_soc_rev2.h
+++ b/mcux/mcux-sdk/devices/MIMX9352/drivers/fsl_edma_soc_rev2.h
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2023 NXP
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef _FSL_EDMA_SOC_REV2_H_
+#define _FSL_EDMA_SOC_REV2_H_
+
+#include "fsl_edma_rev2.h"
+
+#if defined(__cplusplus)
+extern "C" {
+#endif
+
+#define EDMA3_CONFIG (&s_edmaConfigs[0])
+#define EDMA4_CONFIG (&s_edmaConfigs[1])
+
+static const uint32_t s_edma3RegisterLayout[] = {
+    /* MP-related layout */
+    [EDMA_MP_CS_INDEX] = 0x0,
+    [EDMA_MP_ES_INDEX] = 0x4,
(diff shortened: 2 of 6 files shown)


LaurentiuM1234:
@dbaluta @iuliana-prodan could you please help review this as well? Thanks!

mmahadevan108:
@LaurentiuM1234 , the SDK team provide `edma4` as the driver for this EDMA IP version. Can you please pull this in instead of the driver you have created.

mmahadevan108:
cc @DerekSnell

LaurentiuM1234:
MCUX-SDK dependency has been merged, removing from draft.

mcux/hal_nxp.cmake
```diff
@@ -215,6 +215,7 @@ include_driver_ifdef(CONFIG_DMA_MCUX_SMARTDMA		smartdma	driver_lpc_smartdma)
 include_driver_ifdef(CONFIG_DAC_MCUX_LPDAC			dac_1		driver_dac_1)
```
dbaluta, line 215: Commit message says CONFIG_DMA_NXP_IMX_EDMA but the actual code in hal_nxp.cmake uses CONFIG_DMA_NXP_EDMA, we need some clarifications here.

LaurentiuM1234: Fixed. Forgot to change commit message after changing the naming scheme on Zephyr side.

dbaluta (approved):

dbaluta:
@LaurentiuM1234 Looks good to me! thanks Laurentiu!

mcux/mcux-sdk/devices/MIMX9352/drivers/fsl_edma_soc_rev2.h
```diff
@@ -0,0 +1,100 @@
+/*
```
iuliana-prodan, line 1: Can you, please, add a similar file for 8QM?
You said you tested 8QM with EDMA and is ok, right?
IMO is better to added it now since 8QM is also a "user" of revision 2 of the edma driver.

LaurentiuM1234: > You said you tested 8QM with EDMA and is ok, right?

Yes.


> IMO is better to added it now since 8QM is also a "user" of revision 2 of the edma driver.

The main issue I have with this is the fact that it will increase the time until https://github.com/zephyrproject-rtos/zephyr/pull/65671 can be merged, which, in turn, delays future work on the EDMA driver (for instance, I have some QM-related fixes for the Zephyr driver which I can't upstream at the moment since the driver hasn't been merged yet). If it's alright with you I'd like to get the driver out of the way as quickly as possible. This way I can get the fixes upstream and add support for new platforms in parallel.

iuliana-prodan: https://github.com/zephyrproject-rtos/zephyr/pull/65671 this is already dependent on this PR which is not merged, yet.
For QM you'll also have to go through the same cycle - update hal, update Zephyr, etc. 
Maybe, this way we can skip the "update hal" part?
At least, this is what I thought.

LaurentiuM1234: > For QM you'll also have to go through the same cycle - update hal, update Zephyr, etc.

Exactly, but the pipeline also implies updating MCUX-SDK so it would look more like this:

Update MCUX-SDK ---> Update HAL ---> Update Zephyr

This means having to involve another step in this, which is a source of delay. From my previous experience, the main sources of delay are the MCUX-SDK and Zephyr steps. If we were to also add the QM part into this patch we would skip the "Update HAL" step like you said but this step has never been a great source of delay in the pipeline anyways so we're fixing a problem which is not that impactful.

iuliana-prodan: Ok, I understand.
For 8QM you have more files to update/create so you still end up with mcux-sdk->hal->zephyr.

iuliana-prodan (approved):

dbaluta (approved):

dbaluta:
Thanks @LaurentiuM1234. So, dependency https://github.com/nxp-mcuxpresso/mcux-sdk/pull/152 was merged.

@dleach02  this is good to merge!

dleach02 (approved):

dleach02:
Please update your Zephyr PR to point to the SHA

dbaluta:
@dleach02 done. Please help merge this: https://github.com/zephyrproject-rtos/zephyr/pull/65671

Cc: @teburd

PR #308, opened 2023-12-05
s32: drivers: s32ze: extern 'Netc_Eth_Ip_MSIX_Rx' function to use within Netc shimdriver

congnguyenhuu:
Netc shim driver uses 'Netc_Eth_Ip_MSIX_Rx' function. Need to remove 'static inline' keywords from this function so that it can be used as an extern function outside of the file declaration.


s32/drivers/s32ze/Eth_NETC/src/Netc_Eth_Ip_Irq.c
```diff
@@ -108,7 +108,7 @@ static inline void Netc_Eth_Ip_MSIX_Tx(uint8 CtrlIndex);
  *
  * @param CtrlIndex
  */
-static inline void Netc_Eth_Ip_MSIX_Rx(uint8 CtrlIndex);
+void Netc_Eth_Ip_MSIX_Rx(uint8 CtrlIndex);
```
manuargue, line 111: can we avoid this and use `Netc_Eth_Ip_x_MSIX_RxEvent` instead?

manuargue (changes requested):

s32/drivers/s32ze/Eth_NETC/src/Netc_Eth_Ip_Irq.c
congnguyenhuu: we can not use `Netc_Eth_Ip_x_MSIX_RxEvent` when using `DT_INST` for NETC. Because `x` number in `Netc_Eth_Ip_x_MSIX_RxEvent` is the Station Interface (SI) instance index, it not align with the devicetree instance index.

manuargue (approved):

Dat-NguyenDuy (approved):

PR #309, opened 2023-12-05
s32: mcux: s32k146: add support for cache driver

manuargue:
Add definitions to use MCUX SDK cache driver for LMEM controller on S32K146 devices.

Note that `LMEM_PCCCR_ENWRBUF` is not present on S32K1xx devices so it is defined to have no effect if used in the cache driver.


manuargue:
@henrikbrixandersen tagging you in case you want to have a look (it seems that I cannot add you as reviewer on this repo?)

PR #315, opened 2024-01-04
mcux-sdk-middleware-usb: fix ip control return value for ip3511

jsiverskog:
without this kStatus_USB_Error was always returned for kUSB_DeviceControlRun and kUSB_DeviceControlStop control types. let's do it as for ehci/khci.

this change fixes the tests/subsys/usb/device/usb.device test for lpcxpresso55s69_cpu0.

this is from https://github.com/nxp-mcuxpresso/mcux-sdk-middleware-usb/pull/6

### DIFF
--- a/mcux/README
+++ b/mcux/README
@@ -83,3 +83,4 @@ Patch List:
   19. drivers: fsl_sai: Allow NULL FIFO watermark
   20. drivers: fsl_sai: Fix compilation warnings when asserts are disabled
   21. mcux-sdk/drivers/ftm/fsl_ftm.c: make public FTM_GetInstance() function (mcux-sdk#155)
+  22. mcux-sdk-middleware-usb: fix ip control return value for ip3511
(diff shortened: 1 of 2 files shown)


danieldegrasse:
Thanks for this fix- confirmed it fixes the failing test on my end, as well as fixes support for USB DFU mode on the LPC platforms.

A few things:
- Although there is an associated PR to the MCUX USB middleware repository, we should still denote this patch in the patchlist within the README: https://github.com/zephyrproject-rtos/hal_nxp/blob/master/mcux/README#L85
- In order to merge this fix into the HAL, we need to create an associated PR to the Zephyr repository. This PR can simply update the west manifest to point to this PR. See here for some documentation on how to format this: https://docs.zephyrproject.org/latest/develop/modules.html#process-for-submitting-changes-to-existing-modules

jsiverskog:
@danieldegrasse thanks. i have added an entry to the README, and created a corresponding pr here: https://github.com/zephyrproject-rtos/zephyr/pull/67213

mcux/README
```diff
@@ -83,3 +83,4 @@ Patch List:
   19. drivers: fsl_sai: Allow NULL FIFO watermark
   20. drivers: fsl_sai: Fix compilation warnings when asserts are disabled
   21. mcux-sdk/drivers/ftm/fsl_ftm.c: make public FTM_GetInstance() function (mcux-sdk#155)
+  22. mcux-sdk-middleware-usb: fix ip control return value for ip3511
```
danieldegrasse, line 86: Can you make sure there is still a newline at the end of the file here?

jsiverskog: newline added.

danieldegrasse: You might need to check your editor settings- you can also verify that the change looks right with `git diff`. In the initial push, you had added a line here but had not added a newline at the end of the file in this push, you have added an additional newline. Ideally the diff would just be the addition of the line, like so:

image

jsiverskog: ok now? not sure why the newline was removed in the first commit.

danieldegrasse (approved):

mcux/README
danieldegrasse: Yup, looks good- no worries, and thanks :)

PR #323, opened 2024-01-23
mcux: hal_nxp: Remove unneeded files

dleach02:
Remove files not needed by Zephyr to reduce the size of the NXP HAL.


mmahadevan108:
Can we add the folders that were deleted to the README file so these do not get reintroduced by future SDK updates.

dleach02:
> Can we add the folders that were deleted to the README file so these do not get reintroduced by future SDK updates.

done

mmahadevan108:
Or we could reference this PR in the `README`.

mmahadevan108 (approved):

PR #326, opened 2024-01-29
hal_nxp.cmake: fix bad edma directory naming

LaurentiuM1234:
When using the new Zephyr EDMA driver, Zepyhr tries to include the old "edma" directory, which results in a build error since "driver_edma_rev2.cmake" is not placed inside that directory. To fix this, change the included directory's name to edma_rev2, which is the directory that contains the files associated with EDMA rev2 driver from NXP HAL.


LaurentiuM1234:
@mmahadevan108 @dleach02 please review

mmahadevan108 (approved):

PR #332, opened 2024-02-02
Revert "Revert "mcux: hal_nxp: updated cmake to reference new pit definition""

decsny:
Reverts zephyrproject-rtos/hal_nxp#331


danieldegrasse (approved):

decsny:
closing because this is changing a duplicate entry, one of which should be deleted, so no point unreverting this commit

PR #341, opened 2024-02-22
mcux: CMakeLists: update MCUX SOC variable logic for HWMv2

danieldegrasse:
Update MCUX SOC variable name logic to use CONFIG_SOC to determine the MCUX_DEVICE name, and append the core ID based on CONFIG_SOC_XX_CORE Kconfig values.


mcux/hwmv1.cmake
```diff
@@ -0,0 +1,46 @@
+# Get MCUX_DEVICE and MCUX_DEVICE_PATH from CONFIG_SOC
+# For example, if CONFIG_SOC=mimx8ml8_ca53, then MCUX_DEVICE=MIMX8ML8_ca53
+# and MCUX_DEVICE_PATH=MIMX8ML8
+string(FIND ${CONFIG_SOC} "_" pos)
+if(pos GREATER -1)
+  string(REPLACE "_" ";" MCUX_DEVICE_TMP ${CONFIG_SOC})
+  list(GET MCUX_DEVICE_TMP 0 MCUX_DEVICE_SOC)
+  list(GET MCUX_DEVICE_TMP 1 MCUX_DEVICE_CORE)
+  string(TOUPPER ${MCUX_DEVICE_SOC} MCUX_DEVICE_PATH)
+  set(MCUX_DEVICE ${MCUX_DEVICE_PATH}_${MCUX_DEVICE_CORE})
```
JiafeiPan, line 10: we still need these code for hardware model v2, in fact these code can achive what we needed for v2. 
For example, if CONFIG_SOC=mimx8ml8_ca53, then MCUX_DEVICE=MIMX8ML8_ca53 and MCUX_DEVICE_PATH=MIMX8ML8, and MCUX_CPU = CPU_${CONFIG_SOC_PART_NUMBER})
In fact, I still don't clear what's the difference for v2 and v1 you want to update in this PR.

JiafeiPan: I found Cortex-A Core's Part number are with cpu suffix, such as CONFIG_SOC_PART_NUMBER="MIMX8MM6DVTLZ_ca53"
So I will update them to remove cpu suffix. maybe this is one difference for hardware model v1 and v2.

mcux/hwmv2.cmake
```diff
@@ -0,0 +1,41 @@
...
+  set(MCUX_CORE_SUFFIX _cm33_core1)
+elseif(CONFIG_SOC_LPC54114_M4)
+  set(MCUX_CORE_SUFFIX _cm4)
+elseif(CONFIG_SOC_LPC54114_M0)
+  set(MCUX_CORE_SUFFIX _cm0plus)
+elseif(CONFIG_SOC_MIMXRT685S_CM33)
+  set(MCUX_CORE_SUFFIX _cm33)
+elseif(CONFIG_SOC_MIMXRT595S_CM33)
+  set(MCUX_CORE_SUFFIX _cm33)
+endif()
```
JiafeiPan, line 36: For hardware model v2, we will use CONFIG_SOC to figure the CPU on SoC with multiple CPU clusters, for example on imx8mm, use "CONFIG_SOC="mimx8mm6_ca53"" for Cortex A53 Core, and "CONFIG_SOC="mimx8mm6_cm4""  for M4 Core, so we need to add more device list at here, I am thinking whether we could use a script to do it, so that we will not need to update at here if we want to add a new device.

mcux/hwmv2.cmake
```diff
@@ -0,0 +1,41 @@
...
+# MCUX_CPU: "CPU"+ SOC part number, followed by core name when using a dual core part.
+#              Example: CPU_MIMXRT595SFAWC_cm33, or  CPU_LPC55S36JBD100
+# MCU_DEVICE_PATH: SOC name without core suffix. Must match the name of the
+#              folder in MCUX HAL. IE MIMXRT595S, or LPC55S36
+
+
+# For most SOCs, the below settings will suffice
+string(TOUPPER ${CONFIG_SOC} MCUX_DEVICE_PATH)
+string(TOUPPER ${CONFIG_SOC} MCUX_DEVICE)
+set(MCUX_CPU CPU_${CONFIG_SOC_PART_NUMBER})
```
JiafeiPan, line 13: Suggest code here:

string(FIND ${CONFIG_SOC} "_" pos)
if(pos GREATER -1)
  string(REPLACE "_" ";" MCUX_DEVICE_TMP ${CONFIG_SOC})
  list(GET MCUX_DEVICE_TMP 0 MCUX_DEVICE_SOC)
  list(GET MCUX_DEVICE_TMP 1 MCUX_DEVICE_CORE)
  string(TOUPPER ${MCUX_DEVICE_SOC} MCUX_DEVICE_PATH)
  set(MCUX_DEVICE ${MCUX_DEVICE_PATH}_${MCUX_DEVICE_CORE})
  **set(MCUX_CORE_SUFFIX _${MCUX_DEVICE_CORE})**
else()
  string(TOUPPER ${CONFIG_SOC} MCUX_DEVICE)
  string(TOUPPER ${CONFIG_SOC} MCUX_DEVICE_PATH)
endif()
set(MCUX_CPU CPU_${CONFIG_SOC_PART_NUMBER})

mcux/hwmv1.cmake
danieldegrasse: For hardware model V2, the `CONFIG_SOC` value should match the SOC name given in `soc.yml`, which will be `mimx8ml8` for the case you mention. Note that this restriction is being enforced by a CI check upstream (https://github.com/zephyrproject-rtos/zephyr/pull/69309), which is why we are trying to align with it here. Once we remove CPU suffixes, we can use `CONFIG_SOC_xxx_CORE_NAME` Kconfigs, like `CONFIG_SOC_MIMX8ML8_CA53` to set the `MCUX_CORE_SUFFIX` variable

In your case, I would expect the following Kconfig values:
```
CONFIG_SOC_PART_NUMBER=MIMX8MM6DVTLZ
CONFIG_SOC=MIMX8ML8
```

Generally, we will align `CONFIG_SOC` with the folder name used for the SOC in MCUX SDK

mcux/hwmv2.cmake
danieldegrasse: I have wondered about this as well. The core suffix is very much an MCUX specific value, so we can't parse it from `CONFIG_SOC` or something similar. However, we could add a Kconfig such as `CONFIG_MCUX_CORE` in Zephyr, and use that here. Do you think that would make sense?

JiafeiPan: Currently we have "CONFIG_SOC="mimx8mm6_ca53"" and" CONFIG_SOC_PART_NUMBER="MIMX8MM6DVTLZ_ca53", but still have no any configuration item to specify the cpu cluster which is listed in soc.yml, I suggest to add a more common item, just "CONFIG_SOC_CPU",  so that we can align all platforms, what do you think?

danieldegrasse:
@JiafeiPan I have updated this PR to now use a Kconfig, `CONFIG_MCUX_CORE_SUFFIX`. This way, SOCs can set the core suffix in their `Kconfig.defconfig` file.

A few examples of this:
`soc/nxp/imxrt/rt11xx/Kconfig.defconfig`:
```
config MCUX_CORE_SUFFIX
	default "_cm4" if CPU_CORTEX_M4
	default "_cm7" if CPU_CORTEX_M7
```
`soc/nxp/imxrt/rt6xx/Kconfig.defconfig`:
```
config MCUX_CORE_SUFFIX
	default "_cm33" if CPU_CORTEX_M33
```

`soc/nxp/lpc/lpc55xxx/Kconfig.defconfig`:
```
config MCUX_CORE_SUFFIX
	default "_cm33_core0" if SOC_LPC55S69_CPU0
	default "_cm33_core1" if SOC_LPC55S69_CPU1
```

`soc/nxp/lpc/lpc54xxx/Kconfig.defconfig`:
```
config MCUX_CORE_SUFFIX
	default "_cm4" if SOC_LPC54114_M4
	default "_cm0plus" if SOC_LPC54114_M0
```

mcux/hwmv2.cmake
danieldegrasse: We could certainly try to introduce a more common item- the reason I do not want to use a common identifier here is that the strings MCUX SDK uses to identify cores are slightly arbitrary- of course they do describe the core present on the system, but I am sure someone could argue the best string to describe a "Cortex M33" is "M33" rather than "CM33", which is what the HAL uses. A good example of this is the M0+ core on the LPC54114, which is called "cm0plus" by the HAL.

I've updated this PR to use an MCUX specific Kconfig, please let me know your thoughts

mcux/hwmv2.cmake
```diff
@@ -0,0 +1,18 @@
...
+
+if(DEFINED CONFIG_MCUX_CORE_SUFFIX)
+  string(APPEND MCUX_DEVICE ${CONFIG_MCUX_CORE_SUFFIX})
```
JiafeiPan, line 16: As "string(TOUPPER ${CONFIG_SOC} MCUX_DEVICE)", so CONFIG_SOC will not with CPU suffix, right? then CONFIG_SOC will have a single value and use CONFIG_MCUX_CORE_SUFFIX to differentiate CPU Core on this SoC.

JiafeiPan, line 16: As "string(TOUPPER ${CONFIG_SOC} MCUX_DEVICE)", so CONFIG_SOC will not with CPU suffix, right? then CONFIG_SOC will have a single value and use CONFIG_MCUX_CORE_SUFFIX to differentiate CPU Core on this SoC.

JiafeiPan:
> @JiafeiPan I have updated this PR to now use a Kconfig, `CONFIG_MCUX_CORE_SUFFIX`. This way, SOCs can set the core suffix in their `Kconfig.defconfig` file.
> 
> A few examples of this: `soc/nxp/imxrt/rt11xx/Kconfig.defconfig`:
> 
> ```
> config MCUX_CORE_SUFFIX
> 	default "_cm4" if CPU_CORTEX_M4
> 	default "_cm7" if CPU_CORTEX_M7
> ```
> 
> `soc/nxp/imxrt/rt6xx/Kconfig.defconfig`:
> 
> ```
> config MCUX_CORE_SUFFIX
> 	default "_cm33" if CPU_CORTEX_M33
> ```
> 
> `soc/nxp/lpc/lpc55xxx/Kconfig.defconfig`:
> 
> ```
> config MCUX_CORE_SUFFIX
> 	default "_cm33_core0" if SOC_LPC55S69_CPU0
> 	default "_cm33_core1" if SOC_LPC55S69_CPU1
> ```
> 
> `soc/nxp/lpc/lpc54xxx/Kconfig.defconfig`:
> 
> ```
> config MCUX_CORE_SUFFIX
> 	default "_cm4" if SOC_LPC54114_M4
> 	default "_cm0plus" if SOC_LPC54114_M0
> ```



> config MCUX_CORE_SUFFIX

As CONFIG_MCUX_CORE_SUFFIX is SoC related item, not sure it will be better to define it in Kconfig.soc ?

danieldegrasse:
> As CONFIG_MCUX_CORE_SUFFIX is SoC related item, not sure it will be better to define it in Kconfig.soc

I see the argument here. We certainly could place this definition in `Kconfig.soc`. My initial instinct was `Kconfig.defconfig` because `Kconfig.soc` seems to only be for "generic" SOC Kconfigs like `CONFIG_SOC_PART_NUMBER`

JiafeiPan (approved):

mcux/CMakeLists.txt
```diff
@@ -1,52 +1,26 @@
...
-elseif("${MCUX_DEVICE}" STREQUAL "MIMXRT1166_cm4")
-  set(MCUX_CPU CPU_${CONFIG_SOC_PART_NUMBER}_cm4)
-elseif("${MCUX_DEVICE}" STREQUAL "MIMXRT1166_cm7")
-  set(MCUX_CPU CPU_${CONFIG_SOC_PART_NUMBER}_cm7)
-else()
+  string(TOUPPER ${CONFIG_SOC} MCUX_DEVICE)
   set(MCUX_CPU CPU_${CONFIG_SOC_PART_NUMBER})
+
+  if(DEFINED CONFIG_MCUX_CORE_SUFFIX)
+    string(APPEND MCUX_DEVICE ${CONFIG_MCUX_CORE_SUFFIX})
```
LaurentiuM1234, line 21: What is this for? Do we plan to split off the MCU devices by core? (e.g: MIMX8QM6_dsp and MIMX8QM6_ca53?) If so, it should probably be solved in this PR as well as this new approach breaks things. See https://github.com/zephyrproject-rtos/zephyr/pull/69398.

danieldegrasse: Yes, we split off MCUX devices by core currently (within the folder for the SOC). You can see an example of this for the MIMXRT1176 SOC: https://github.com/zephyrproject-rtos/hal_nxp/tree/master/mcux/mcux-sdk/devices/MIMXRT1176

PR #343, opened 2024-03-01
Compile in PXP driver when LVGL is set to use PXP

dberlin:
This is part of adding support for LVGL to use PXP under zephyr.

(There are LVGL patches being submitted as well)

ELCDIF accelerated rotation exists in zephyr, using a PXP DMA driver, but it is not
possible to reuse that support for zephyr PXP support in LVGL.

In practice, LVGL wants to be able to tell when the PXP is idle or not, in a thread-safe way.
Zephyr considers DMA to be single-owner and explicitly does not want these sorts of synchronization primitives
in the engine, etc. So the Zephyr PXP DMA driver does not do any locking.


mmahadevan108:
Would'nt `CONFIG_MCUX_PXP` be enabled when LVGL is used with PXP?

danieldegrasse:
> (There are LVGL patches being submitted as well)

Can you provide a link to these patches? Interested to see what is being implemented. Is the support for the PXP being implemented entirely outside the Zephyr tree, or will a minimal driver be implemented in the Zephyr tree?

dberlin:
> Would'nt `CONFIG_MCUX_PXP` be enabled when LVGL is used with PXP?
That is what this patch does.

Right now, it is only enabled if you select the DMA driver for pxp + ELCDIF driver.

Neither of which is necessary (or actually useful) for LVGL PXP support.


This patch makes it so if you select the LVGL PXP option, it also enables CONFIG_MCUX_PXP regardless of the state of the zephyr dma driver and ELCDIF driver.

danieldegrasse:
> Neither of which is necessary (or actually useful) for LVGL PXP support.

Yup, the DMA driver in tree is really for a different use case (statically rotating/converting color space of the display), whereas the PXP itself has quite a few more blit capabilities.

dberlin:
> > (There are LVGL patches being submitted as well)
> 
> Can you provide a link to these patches? Interested to see what is being implemented. Is the support for the PXP being implemented entirely outside the Zephyr tree, or will a minimal driver be implemented in the Zephyr tree?

Sure.
https://github.com/zephyrproject-rtos/lvgl/compare/zephyr...dberlin:lvgl:zephyr

It's not a lot, as you can see.

This is just me (IE i have no affiliation with anything).

For the purposes of LVGL, *as it exists right now*,  no in-tree *zephyr* changes are necessary beyond 

1. enabling the MCUX PXP driver if you select the right LVGL option.

2. Defining the display height/width and OS_SDK_ZEPHYR (lvgl relies on OS_SDK_FREERTOS) in lvgl_support.h.
LVGL's PXP support currently expects a header file with those things in it.

Within LVGL, it is also a very small set of changes.

Why: LVGL's current PXP support expects to handles everything from clock init to you name it. 

If you look at https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/dma/dma_mcux_pxp.c , LVGL's current PXP support expects to do everything you see in dma_mcux_pxp_init, itself.

See here: https://github.com/dberlin/lvgl/blob/zephyr/src/draw/nxp/pxp/lv_gpu_nxp_pxp.c
look at the init function.

This could be changed to rely on zephyr to do it, and we could create a driver that does it, but right now, the pxp support in LVGL is lower level and expects exclusivity.

This, of course, causes the zephyr dma/elcdif drivers to conflict with LVGL PXP support.

If we want them not to conflict, it would not be enough to create a pxp driver, we'd need to abstract out PXP from the ELCDIF/DMA parts so that, for example, submit/completion state can be given to LVGL and not conflict with it or ELCDIF or ...

(Also, right now, zephyr's PXP accelerated ELCDIF driver  and LVGL both create a framebuffer for PXP)

So i'll make the following offer:
If someone comes up with an agreed-upon (between lvgl/zephyr folks) plan of how to make this all work better i'm happy to code the pieces.

I know that tends to take a lot of time and thought, and I don't have a strong desire to do it since it's a hobby for me. 

The coding part and shepherding part is not that hard here, so i'm happy to do that piece.

Note: I have also added vglite zephyr support (but have not fully tested it yet).  This is less problematic because Zephyr does not conflict it.

danieldegrasse:
> I know that tends to take a lot of time and thought, and I don't have a strong desire to do it since it's a hobby for me.

> The coding part and shepherding part is not that hard here, so i'm happy to do that piece.

> Note: I have also added vglite zephyr support (but have not fully tested it yet). This is less problematic because Zephyr does not conflict it.

> You can find the rough version here:
> https://github.com/dberlin/vglite_zephyr

> and a west manifest that uses it here:
> https://github.com/dberlin/zephyr/blob/main/west.yml

> This will not by itself work, i have not finished adding a vglite HAL port that does the clock/etc init for zephyr (in FreeRTOS the > apps do it).

> Once i get it finished up i will probably try to make a zephyr vglite driver and move the HAL pieces there.

Very cool to see! I have done some work on enabling VGLite with Zephyr as well. Thanks for putting this out, I'm going to see if I can get that work moving forwards. The main obstacle I recall is that I'm not aware of a public release of VGLite currently on github (unless there is one I'm missing)...

danieldegrasse (approved):

dberlin:
> > I know that tends to take a lot of time and thought, and I don't have a strong desire to do it since it's a hobby for me.
> 
> > The coding part and shepherding part is not that hard here, so i'm happy to do that piece.
> 
> > Note: I have also added vglite zephyr support (but have not fully tested it yet). This is less problematic because Zephyr does not conflict it.
> 
> > You can find the rough version here:
> > https://github.com/dberlin/vglite_zephyr
> 
> > and a west manifest that uses it here:
> > https://github.com/dberlin/zephyr/blob/main/west.yml
> 
> > This will not by itself work, i have not finished adding a vglite HAL port that does the clock/etc init for zephyr (in FreeRTOS the > apps do it).
> 
> > Once i get it finished up i will probably try to make a zephyr vglite driver and move the HAL pieces there.
> 
> Very cool to see! I have done some work on enabling VGLite with Zephyr as well. Thanks for putting this out, I'm going to see if I can get that work moving forwards. The main obstacle I recall is that I'm not aware of a public release of VGLite currently on github (unless there is one I'm missing)...

Yes, that is totally correct.
The code here comes from an mcuxpresso download.  

I was actually going to poke NXP and see if they were willing to just throw it on github, with the expectation that it might take a while.   I doubt it is deliberate strategy on their part as much as haven't gotten around to it/nobody asked us to.

I also used to run open source licensing for Google (and have a law degree) so i'm also pretty good at speaking lawyerese if it becomes necessary ;)

dberlin:
Happy to see it happen!

The last piece you need to make it all work is this:
https://github.com/zephyrproject-rtos/zephyr/commit/dccd4f0f7c46e2a46d54d0791bb353a7d8f84a25

With that in place it works and i've run it multi-threaded pretty heavily
with no deadlocks/etc.

The one issue i have not tracked down yet is that I get BGR vs RGB
inversion on some panels i tried.

(IE with PXP deep red is deep blue, without PXP deep red is deep red)

I know NXP did updates to PXP support upstream as part of 9.x, and I
haven't looked to see if that fixes it.

--Dan

danieldegrasse:
> The last piece you need to make it all work is this:
https://github.com/zephyrproject-rtos/zephyr/commit/dccd4f0f7c46e2a46d54d0791bb353a7d8f84a25

Was wondering how we were handling this. This is the one tricky bit for the VGLite and PXP- we can implement *most* of the driver outside of Zephyr, but things like interrupt connection (and reading base address) sometimes require code on the Zephyr side. Not sure how we best manage this, but I appreciate the effort here!

dleach02 (approved):

mmahadevan108 (approved):

mmahadevan108:
This PR has been merged. Is there a manifest file that needs to be updated to get this change?

PR #348, opened 2024-03-07
mcux: hal_nxp: Include mipi_csi2rx driver

ngphibang:
This is to include the low level mipi csi-2 Rx driver which is needed for [this PR](https://github.com/zephyrproject-rtos/zephyr/pull/69810)


decsny:
This is a bit of a nitpick but it's a bit unconventional to have the vendor name at the end of a kconfig string, can you change it to CONFIG_MCUX_MIPI_CSI2RX

ngphibang:
@decsny Thanks. Yes, I think so, too. I hesitated so much about the naming. Frankly, I am a bit confused between "mcux" and "imx" (imx seems covers both mcu and mpu, etc). As this is also part of the[ zephyr PR ](https://github.com/zephyrproject-rtos/zephyr/pull/69810), could we discuss there so that I could have more advice from others than I will change the naming all at once ?

ngphibang:
@decsny I changed the config name.

ngphibang:
@decsny Could you approve and merge the PR ?
Thanks !

decsny (approved):

PR #366, opened 2024-04-03
mcux-sdk: drivers: csi: Rework to fix framerate issue

ngphibang:
Currently, the camera framerate is divided by two. This is because the driver only puts the full buffer to the queue and calls the callback at half of the interrupt rate (when there are at least 2 active framebuffers).

On **bare metal**, this issue is not really visible when using more than 3 buffers and the application does not hold the full buffer for too long before submitting it again to the empty buffers queue. But if we use only 2 buffers, the framerate is clearly reduced (by a factor of two).

Especially, when CSI_TransferGetFullBuffer() get called inside the callback function **(as in Zephyr RTOS), framerate is always reduced by a factor of two despite of the number of buffers used.**

Rework the driver to fix these issues as well as to clean up the logics.

The corresponding SDK PR is here : https://github.com/nxp-mcuxpresso/mcux-sdk/pull/188

### DIFF
--- a/mcux/README
+++ b/mcux/README
@@ -77,3 +77,4 @@ Patch List:
   10. devices: LPC5534: Changed FSL_FEATURE_PWM_HAS_NO_WAITEN from 1 to 0 to solve pwm issue.
   11. devices: LPC5536: Changed FSL_FEATURE_PWM_HAS_NO_WAITEN from 1 to 0 to solve pwm issue.
   12. devices: LPC55S36: Changed FSL_FEATURE_PWM_HAS_NO_WAITEN from 1 to 0 to solve pwm issue.
+  13. devices: MIMX8UD7: add definition for LPUART_RX_TX_IRQS


mmahadevan108:
Please fix the merge conflict

ngphibang:
Yes, here it is : https://github.com/zephyrproject-rtos/zephyr/pull/71346

mmahadevan108:
Please address merge conflict and also latest updates to the changes to the SDK driver per feedback you received in your PR to the SDK driver.

ngphibang:
Hi @mmahadevan108,

- I already added the lastest code addressed the feedback from SDK PR
- There is merge conflict in README on the patch list because there is always new patches merged everyday. I suggest to resolve this only when this PR is on the point to be merged (I cannot resolve it everyday) ?

ngphibang:
@mmahadevan108
Now, the PR contains 0 commits and is closed.
What I have done is I just resolved the conflict directly from the github website but then I realized that it's not the good way to do. So I resolved it in my local and pushed force but after that, the PR is closed (??) and I cannot reopen it.

Do you know why ? Should I create another one ?

PR #372, opened 2024-04-12
drivers: csi: Rework to fix the low framerate issue

ngphibang:
Currently, the camera framerate is divided by two. This is because the driver only puts the full buffer to the queue and calls the callback at half of the interrupt rate (when there are at least 2 active framebuffers).

This issue is not really visible when using more than two buffers and the application does not hold the full buffer for too long before submitting it again to the empty buffers queue. But if we use only two buffers, the framerate is clearly reduced (by a factor of two).

Especially, when CSI_TransferGetFullBuffer() get called inside the callback function (as in Zephyr RTOS), framerate is always reduced by a factor of two despite of the number of buffers used.

Rework the driver to fix these issues as well as to clean up the logics.

The corresponding SDK PR is here https://github.com/nxp-mcuxpresso/mcux-sdk/pull/188

- I already updated this PR with the **latest code from the SDK PR** which addressed the comments from the reviewer.
- As the patch list in the README is changed nearly everyday, this PR may have a merge conflict (on the patch number in README). I suggest to solve this merge conflict when the PR is on the point to be merged.

### DIFF
--- a/mcux/README
+++ b/mcux/README
@@ -75,3 +75,4 @@ Patch List:
       - Update usb_device_ehci.c and usb_device_lpcip3511.c if in Zephyr environment.
         - Add include of usb_device_mcux_drv_port.h.
         - Remove include of usb_device.h and usb_device_dci.h.
+  7. drivers: csi: Rework to fix the low framerate issue
--- a/mcux/mcux-sdk/drivers/csi/fsl_csi.c
+++ b/mcux/mcux-sdk/drivers/csi/fsl_csi.c
@@ -65,56 +65,6 @@ typedef union pvoid_to_u32
 static uint32_t CSI_GetInstance(CSI_Type *base);
 
 #if !CSI_DRIVER_FRAG_MODE
-/*!
- * @brief Get the delta value of two index in queue.
- *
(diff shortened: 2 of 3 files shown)


dleach02:
@ngphibang, I think we need to wait on your SDK PR to resolve its discussions before merging this one.

dleach02 (changes requested):
Wait on SDK PR https://github.com/nxp-mcuxpresso/mcux-sdk/pull/188 to resolve discussions.

ngphibang:
@dleach02 @mmahadevan108 : I had discussions with Jason (the PR is technically correct. The discussions are just about keeping or not the exact original behavior of calling the callback in case of someone has been familiar with the old behavior, if any) and we come to an agreement.
The PR is now merged into MCUX SDK. Could you help to merge this into the hal_nxp ?
Thanks !

dleach02 (approved):

dleach02:
@mmahadevan108

ngphibang:
Just resolved the merge conflict of the patch number in README as this is on the point to be merged

decsny (approved):

PR #382, opened 2024-05-01
Added support for S32K142 SoC

vChavezB:
This PR addreses the S32K142 SoC. My motivation was that I need to develop a project with a custom board and I thought why not use Zephyr instead. The changes are based on the S32K146 SoC, thanks a lot for that ! 

So far I have tested FTM, FlexCAN, and GPIOs. 

Unfortunately I do not have a board definition as I cannot release what I am working on. Hence the Clock configuration is based on an external 40 MHz clock (SPLL).

In addition had to delete some definitions and manually make sure that by deleting definitions the Look up tables for clock tables,etc. work.

Hope it helps !


vChavezB:
I have tested the FlexNVM as EEPROM in a separate branch here

https://github.com/vChavezB/hal_nxp/tree/s32k142_flash

The reason I did not add it in this PR is because I find it kind of hacky. I found out after digging the SDK drivers for the S32K142 that there are some features that are not part of this SOC, e.g. Read resources command. So I had to add some ugly ifdef macros in the flash drivers. As I am not an expert in the S32K1 family, I just give this as a reference in case NXP wants to correctly make the flash drivers portable as I do not know which other SoCs from the S32 family have these sort of compatibility problems.

I also made a driver with the zephyr eeprom api. Its a quick and dirty solution hence I do not publish it. I see there is a lot of work to be done with conditional Macros and binding to KConfig to correctly set the EEPROM Size,etc. 

Feel free to use this code as inspiration to add support for EEPROM (FlexNVM) with the S32K1 SoCs.

vChavezB:
Hi @manuargue, thanks for the feedback. Yes it would be better to have a definition for the dev kit to test the changes.

Adding the board definition should not be complicated but I do not have one at hand as I am working with custom boards. If I have some spare time I will try to make a PR in zephyr upstream.

Just as additional comment it would be nice to have integration of the zephyr flash and eeprom api (i.e. PFLASH, FlexNVM).

PR #385, opened 2024-05-06
mcux: remove LA_OPT license files and MCUX SDK SCR

danieldegrasse:
Remove LA_OPT license files from the MCUX SDK, as the MCUX SDK imported into the Zephyr MCUX HAL should not include LA_OPT licensed components. Also, remove the software content register file in the MCUX SDK copy, since the Zephyr MCUX HAL has its own SCR file.

Future HAL updates should review the contents of the MCUX SDK SCR file to insure no new LA_OPT licensed components are added to the Zephyr MCUX SDK.


danieldegrasse:
@yvanderv interested to hear your thoughts on this change- is this the right way to remove these LA_OPT files?

PR #390, opened 2024-05-16
Fix mcux SDK cache64 driver

mmahadevan108:
These changes have been submitted to the SDK team.

### DIFF
--- a/mcux/mcux-sdk/drivers/cache/cache64/fsl_cache.h
+++ b/mcux/mcux-sdk/drivers/cache/cache64/fsl_cache.h
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2021, 2023 NXP
+ * Copyright 2016-2021, 2023-2024 NXP
  * All rights reserved.
  *
  * SPDX-License-Identifier: BSD-3-Clause
@@ -22,7 +22,7 @@
 /*! @name Driver version */
 /*! @{ */
 /*! @brief cache driver version. */
-#define FSL_CACHE_DRIVER_VERSION (MAKE_VERSION(2, 0, 7))
+#define FSL_CACHE_DRIVER_VERSION (MAKE_VERSION(2, 0, 8))
 /*! @} */
 
 /*! @brief cache line size. */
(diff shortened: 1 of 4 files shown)


mmahadevan108:
cc @DerekSnell

mcux/mcux-sdk/drivers/cache/cache64/fsl_cache.h
```diff
@@ -79,12 +79,13 @@ uint32_t CACHE64_GetInstance(CACHE64_POLSEL_Type *base);
...
 /*!
- * brief Returns an instance number given physical memory address.
+ * brief Returns an instance numbers given physical memory address.
  *
  * param address The physical memory address.
+ * param pointer to store the index of the physical memory index
  * @return CACHE64_CTRL instance number starting from 0.
  */
-uint32_t CACHE64_GetInstanceByAddr(uint32_t address);
+uint32_t CACHE64_GetInstanceByAddr(uint32_t address, uint8_t *bufIndex);
```
decsny, line 88: why is the api changing?

mmahadevan108: This is an internal function and not an API change. This function was not functioning as intended. This function was returning a value which was used as an index for 2 arrays.

dleach02: if it is an internal function, then why is it defined in the header file? 

Is the SDK team okay with this change?

mmahadevan108: Working with the SDK team on this change.

dleach02:
This needs to wait on the SDK 2.16.00 PR or these changes will be overwritten.

decsny:
there is a merge conflict

dleach02 (changes requested):
Need to add these patches to the patch list document

mmahadevan108:
These patches have been added into the SDK repo and should be in our HAL when we do the next update. Should I still update the patches list?

mmahadevan108:
@dleach02 PR updated with update to README. Added a patch that has been submitted to the SDK team after testing their new CACHE64 driver.

EmilioCBen (approved):

danieldegrasse (approved):
Just to make sure I understand this change correctly, the CACHE64 invalidation only works correctly if a nonsecure address is written to `CSAR`?

mmahadevan108:
Thats correct.

PR #396, opened 2024-05-25
mcux: hal_nxp: Add QTMR PWM driver

PetervdPerk-NXP:
Adds QTMR based PWM driver.


bperseghetti (approved):
LGTM. Important for VMU_RT1170

PR #398, opened 2024-05-28
Support monolithic app

yeaissa:
Make use of generate_inc_file_for_target() to convert the firmware binaries into C arrays which will be linked by the application and sourced when loading the CPU1 and CPU2 firmwares.

Put the firmwares into sections to allow for more control on placement.

### DIFF
--- a/zephyr/src/rw61x/rw61x_cpu1.c
+++ b/zephyr/src/rw61x/rw61x_cpu1.c
@@ -1,0 +1,20 @@
+/*
+ * Copyright 2023-2024 NXP
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ *
+ */
+
+
+#include <stdint.h>
+
+#if defined(CONFIG_NXP_MONOLITHIC_WIFI)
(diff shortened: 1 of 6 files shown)


zephyr/src/rw61x/rw61x_cpu1.c
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright 2023 NXP
```
axelnxp, line 2: should it be 2023-2024 ?

dleach02 (approved):

decsny:
What's the reasoning for putting this in the HAL instead of in the zephyr tree?

axelnxp:
> What's the reasoning for putting this in the HAL instead of in the zephyr tree? Even you're adding a new folder called zephyr/src/ which didn't exist before

Back when we discussed the binary blob topics (mainly with David), we saw the other vendors such as infineon and esp did it this way, the blobs are fetched and placed into the HAL, and you have the .c file around the same place so everything is centralized.
Also, the part which is searching for the binary is specific to our HAL, so I guess this shouldn't go somewhere else.

PR #411, opened 2024-07-18
s32: convert SIUL2 drivers to native drivers

manuargue:
Convert SIUL2 pin control, GPIO and external interrupt controller to native drivers and remove sources no longer needed.


bperseghetti (approved):

Dat-NguyenDuy (changes requested):
Pls re-generate the new pinmux macros in dts/nxp/s32/*. Previously, we use the RTD driver, it handles the imcr index arbitrarily (i.e on s32z, with siul2_3: imcr index= imcr index in iomux sheet, whether for the rest of instance, imcr index = imcr index in iomux sheet - 512).

With this native approach, i believe this can be done consistently for all siul2 instances

PetervdPerk-NXP (approved):

bperseghetti (approved):

dleach02 (approved):

Dat-NguyenDuy:
> Pls re-generate the new pinmux macros in dts/nxp/s32/*. Previously, we use the RTD driver, it handles the imcr index arbitrarily (i.e on s32z, with siul2_3: imcr index= imcr index in iomux sheet, whether for the rest of instance, imcr index = imcr index in iomux sheet - 512).
> 
> With this native approach, i believe this can be done consistently for all siul2 instances

Thanks for the update, but i have just found one more issue that is on such platform like the s32z, imcr and mscr index for pin may not belong to the same siul2 instance. For e.g PA5_CAN_0_RX: mscr_5-siul2_0; imcr_0 - siul2_3. So i think you might need to extend the macro in https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/dt-bindings/pinctrl/nxp-s32-pinctrl.h#L72 to separate siul2 index to siul2_mscr and siul2_imcr? Your implementation for the pinctrl driver need to be updated also

manuargue:
> Thanks for the update, but i have just found one more issue that is on such platform like the s32z, imcr and mscr index for pin may not belong to the same siul2 instance. For e.g PA5_CAN_0_RX: mscr_5-siul2_0; imcr_0 - siul2_3. So i think you might need to extend the macro in https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/dt-bindings/pinctrl/nxp-s32-pinctrl.h#L72 to separate siul2 index to siul2_mscr and siul2_imcr? Your implementation for the pinctrl driver need to be updated also

please log an issue and let's address that once this pr is merged, because that's a pre-existing issue not related to this pr

Dat-NguyenDuy (approved):

PR #415, opened 2024-08-01
mcux-sdk: update to SDK 2.16.000

dleach02:
Updating mcux components to 2.16.100

Signed-off-by: David Leach <david.leach@nxp.com>

PR #417, opened 2024-08-05
Feature/enable udc sof

MarkWangChinese:
@dleach02 This pr need waiting your SDK 2.16.0 pr (https://github.com/zephyrproject-rtos/zephyr/pull/76604) merged because it depend on the usb controller drivers' updating of SDK 2.16.0.


dleach02:
@MarkWangChinese could we move these changes to the 2.16.000 HAL PR and then update your Zephyr Side PR to point to the HAL 2.16 PR? This way we only have one HAL PR update.

MarkWangChinese:
> @MarkWangChinese could we move these changes to the 2.16.000 HAL PR and then update your Zephyr Side PR to point to the HAL 2.16 PR? This way we only have one HAL PR update.

Hi @dleach02 Thanks. I think it is OK to move these changes to 2.16.000 HAL PR. You can do it directly if you have time. Otherwise, please tell me how to do it.

dleach02:
@MarkWangChinese I merged your changes into #415 PR.

PR #423, opened 2024-08-19
mcux: hal_nxp.cmake: update include fsl_acmp driver ifdef

bjarki-andreasen:
The addition of the comparator subsystem https://github.com/zephyrproject-rtos/zephyr/pull/76087 into zephyr adds a dedicated comparator device driver alongside the existing sensor subsystem based device driver. This causes namespace clashes as the ACMP will now be supported by both a sensor API based driver and comparator based driver.

The kconfig symbol `MCUX_ACMP` will be split into `SENSOR_MCUX_ACMP` and `COMPARATOR_MCUX_ACMP`, both of which require the fsl_acmp driver to be included in the build.

The simplest way of doing this is to add two entries, since both of these symbols will not be selected at the same time.

We could also leave the `CONFIG_MCUX_ACMP` in there as well and have three entries, and then remove `CONFIG_MCUX_ACMP` after https://github.com/zephyrproject-rtos/zephyr/pull/76087 has been merged, that way we don't break anything :)


decsny:
can't we have CONFIG_MCUX_ACMP represent building the hal driver, and the two new symbols selecting it

bjarki-andreasen:
> can't we have CONFIG_MCUX_ACMP represent building the hal driver, and the two new symbols selecting it

That would be neat I think, I can add
```
config MCUX_ACMP
        bool "Include MCUX_ACMP driver"
```

to https://github.com/zephyrproject-rtos/zephyr/blob/main/modules/hal_nxp/Kconfig

and select that from the in-tree drivers :)

That is the pattern we use for hal_nordic as well

decsny:
> > can't we have CONFIG_MCUX_ACMP represent building the hal driver, and the two new symbols selecting it
> 
> That would be neat I think, I can add
> 
> ```
> config MCUX_ACMP
>         bool "Include MCUX_ACMP driver"
> ```
> 
> to https://github.com/zephyrproject-rtos/zephyr/blob/main/modules/hal_nxp/Kconfig
> 
> and select that from the in-tree drivers, then we don't need to change the NXP HAL at all :)
> 
> That is the pattern we use for hal_nordic as well.

yes, we don't do it yet, but I think eventually we should have kconfigs that correspond to including nxp hal drivers. I think in the near future we will be restructuring the NXP HAL significantly and probably this will end up happening for everything

bjarki-andreasen:
> > > can't we have CONFIG_MCUX_ACMP represent building the hal driver, and the two new symbols selecting it
> > 
> > 
> > That would be neat I think, I can add
> > ```
> > config MCUX_ACMP
> >         bool "Include MCUX_ACMP driver"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > to https://github.com/zephyrproject-rtos/zephyr/blob/main/modules/hal_nxp/Kconfig
> > and select that from the in-tree drivers, then we don't need to change the NXP HAL at all :)
> > That is the pattern we use for hal_nordic as well.
> 
> yes, we don't do it yet, but I think eventually we should have kconfigs that correspond to including nxp hal drivers. I think in the near future we will be restructuring the NXP HAL significantly and probably this will end up happening for everything

I updated the comparator PR to add the config and not touch the west manifest for nxp_hal, relevant commit is here https://github.com/zephyrproject-rtos/zephyr/pull/76087/commits/f68292907877ff61515eb21bd5ecd43f141a35ec#diff-bfddd23fc41710fda8969151e660a77adf7b71c7ee6dad94a9f7770cf8e754d5

PR #454, opened 2024-10-18
zephyr: enable CONFIG_NXP_WIFI_BUILD_ONLY_MODE

danieldegrasse:
Add support for CONFIG_NXP_WIFI_BUILD_ONLY_MODE, which enables the NXP WiFi driver to be built by CI without binary blobs


mcux/middleware/wifi_nxp/wifidriver/wpa_supp_if/rtos_wpa_supp_if.c
```diff
@@ -2669,25 +2669,6 @@ void wifi_nxp_wpa_supp_event_proc_dfs_cac_finished(void *if_priv, nxp_wifi_dfs_c
     }
 }
 
-void wifi_nxp_wpa_supp_event_signal_change(void *if_priv, t_s16 *curr_rssi)
```
MaochenWang1, line 2672: The related hostap PR is still under review, which can fix the build error:
https://github.com/zephyrproject-rtos/hostap/pull/38

danieldegrasse: Thanks for the heads up- if the support for 11kvr roaming merges in HostAP first, we can drop this change. I just want to get the upstream WiFi support to a point where it builds and works correctly.

dleach02 (approved):

PR #456, opened 2024-10-24
mcux: hal_nxp: Include SDMA HAL driver if zephyr SDMA driver is enabled

dbaluta:
Zephyr SDMA drivers uses HAL SDMA API so we need to include it.


dbaluta:
Replaced by: https://github.com/zephyrproject-rtos/hal_nxp/pull/458

PR #458, opened 2024-11-05
Include SDMA and PDM MICFIL HAL drivers if enabled from Zephyr

dbaluta:
This will pull in NXP HAL drivers for MICFIL and SDMA.


mmahadevan108:
Is there a Zephyr side PR that needs this change?

mmahadevan108 (approved):

PR #461, opened 2024-11-07
mcux: mcux-sdk: drivers: cache64: remove invalid assert in GetInstance

danieldegrasse:
CACHE64_GetInstanceByAddr() asserts when an address is passed in that is not managed by the CACHE64 controller. This is incorrect behavior, as the CACHE64 driver checks to see if the returned instance number is out of range when using this function, so we should not assert here.

Remove the assert so that cache invalidation functions can be used with addresses not the CACHE64 controller's range.

### DIFF
--- a/mcux/README
+++ b/mcux/README
@@ -121,3 +121,6 @@ Patch List:
   - mcux-sdk/drivers/dspi/fsl_dspi.c, mcux-sdk/drivers/dspi/fsl_dspi.h, mcux-sdk/drivers/dspi/fsl_dspi_edma.c,
   mcux-sdk/drivers/dspi/fsl_dspi_edma.h: add the guards for unsupport features on S32Z27x devices
   - drivers: irqsteer: adjust CHn_MASK index computation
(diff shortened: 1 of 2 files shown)


danieldegrasse:
This change has already been made within the SDK internally, so we should be able to drop this patch by the next SDK release

mcux/README
```diff
@@ -121,3 +121,6 @@ Patch List:
   - mcux-sdk/drivers/dspi/fsl_dspi.c, mcux-sdk/drivers/dspi/fsl_dspi.h, mcux-sdk/drivers/dspi/fsl_dspi_edma.c,
   mcux-sdk/drivers/dspi/fsl_dspi_edma.h: add the guards for unsupport features on S32Z27x devices
   - drivers: irqsteer: adjust CHn_MASK index computation
+  - mcux-sdk/drivers/cache/cache64/fsl_cache.c: remove invalid assertion when
```
decsny, line 124: if it's already in the next sdk, do we need to update the readme still

dleach02: yes

dleach02 (approved):

mmahadevan108 (approved):

PR #465, opened 2024-11-09
added support for S32K148

marcoe87:
added support for S32K148 (umbrella device of S32K1 family)

Note: Clock_Ip_Cfg.c not delivered here since it is board-specific (e.g. installed crystal frequency may change) and IMHO should not be included in HALs. What do you think about this, also regarding other pre-existent S32K1 SOCs?


mmahadevan108:
Is there a Zephyr side PR for this?

dleach02 (changes requested):
The commits here need attribution and need to be signed.

manuargue:
I'd suggest this to be closed in favor of https://github.com/zephyrproject-rtos/hal_nxp/pull/509, that includes all the pieces missing to bringup support for S32K148

PR #468, opened 2024-11-15
 dts: nxp: add pin control files for subset parts

danieldegrasse:
Add pin control files for subset parts within NXP HAL. Note that this
does not add files for all parts supported by the NXP HAL, only for
those that are already supported by the pin control tooling. This
excludes a few part families:
- MKE0* parts
- LPC8* parts
- LPC51* parts
- MIMX8D*, MIMX8Q*, MIMX8U* parts

Note- existing pin control files have not been regenerated, this file only adds new ones


dleach02 (approved):

EmilioCBen (approved):

decsny (approved):

decsny:
Everyone seems to be speaking about this PR as if it already merged and it doesn't affect anything on the zephyr side so I'm just going to merge it

PR #490, opened 2024-12-17
drivers: sai: the 1st frame synchronization signal lost in slave side

Raymond0225:
According the RM document, RT1170 58.3.3 (should be same for other MCU which has a similiar SAI IP):
"A valid frame sync is also ignored (slave mode) or not generated (master mode) for the first four bit clock cycles after enabling the transmitter or receiver."
but in fact, we found master side send out a valid frame sync at the 3rd bit clock cycles which cause this frame sync is ignored by the slave side and frame data lost.
To workaround this issue, bit clock is enabled before TE/RE.

### DIFF
--- a/mcux/mcux-sdk/drivers/sai/fsl_sai.c
+++ b/mcux/mcux-sdk/drivers/sai/fsl_sai.c
@@ -509,8 +509,15 @@ void SAI_TxEnable(I2S_Type *base, bool enable)
         /* If clock is sync with Rx, should enable RE bit. */
         if (((base->TCR2 & I2S_TCR2_SYNC_MASK) >> I2S_TCR2_SYNC_SHIFT) == 0x1U)
         {
+            base->RCSR = ((base->RCSR & 0xFFE3FFFFU) | I2S_RCSR_BCE_MASK);
             base->RCSR = ((base->RCSR & 0xFFE3FFFFU) | I2S_RCSR_RE_MASK);
         }
+        /* Sometimes, bit clock starts just 3 clocks before frame synchronization
+         * signal, which cause the 1st frame sync is ignored by the RX side as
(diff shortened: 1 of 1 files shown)


decsny:
Has this change been submitted to the SDK ? All PRs to this repo that change SDK code need to be upstreamed to the SDK so that we don't need to keep repatching after updating to new SDKs (this repo is a "public downstream" of the "internal upstream" SDK)

Raymond0225:
> Has this change been submitted to the SDK ? All PRs to this repo that change SDK code need to be upstreamed to the SDK so that we don't need to keep repatching after updating to new SDKs (this repo is a "public downstream" of the "internal upstream" SDK)

I am submitting to internal upstream at the same time.

Raymond0225:
This patch has been submitted to internal SDK repo and got merged.

decsny (approved):

EmilioCBen (approved):

dleach02 (approved):

dleach02:
there needs to be a Zephyr side PR to update the manifest. I see that it has been mentioned though in a couple.

dleach02 (changes requested):
Add Zephyr side manifest update

Raymond0225:
> Add Zephyr side manifest update

it is here:  https://github.com/zephyrproject-rtos/zephyr/pull/83718

dleach02 (approved):

Raymond0225:
do you think I should update the readme file and include the description of this patch? 
https://github.com/zephyrproject-rtos/hal_nxp/blob/master/mcux/README#L127

decsny:
> do you think I should update the readme file and include the description of this patch? https://github.com/zephyrproject-rtos/hal_nxp/blob/master/mcux/README#L127

if it was merged to sdk upstream then not needed

PR #504, opened 2025-02-01
cmake: remove shim to lpspi driver

decsny:
Zephyr driver is now native, no need for shim to hal driver.


dleach02:
@decsny please realign with the recent SDK-NG changes. I actually believe this PR can be closed as the file you are touching basically moved to the Zephyr side.

PR #521, opened 2025-03-13
mcux-sdk: drivers: netc: do not limit using ring0 in EP driver

yangbolu1991:
In SDK NETC EP driver, ENETC pseudo MAC ring0 is reserved for switch driver to use. However in zephyr, switch is using DSA driver which requires unmodified host ethernet driver, but not using hal switch driver. So, remove the limitation of using ring0 in EP driver.

Also added the patch to patch list.

### DIFF
--- a/mcux/README
+++ b/mcux/README
@@ -192,3 +192,4 @@ Patch List:
   - mcux-sdk-middleware-ieee_802.15.4: adapt code to be supported with Zephyr
   - mcux-sdk-middleware-usb: Disable the control endpoint ZLT (Zero Length Transaction) of MCUX hal USB controller drivers
     (usb_device_ehci.c and usb_device_lpcip3511.c) in Zephyr environment.
+  - mcux-sdk-ng: drivers: netc: do not limit using ring0 in EP driver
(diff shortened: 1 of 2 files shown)


dleach02:
@yangbolu1991 has this change been submitted back to the SDK-NG?

yangbolu1991:
@dleach02 
As the description in commit message, this patch is only for zephyr.
Thanks.

mmahadevan108:
> @dleach02 As the description in commit message, this patch is only for zephyr. Thanks.

@yangbolu1991, this is not a good long term solution. We need to avoid maintaining SDK driver patches in the Zephyr HAL as it complicates updates of SDK drivers.

butok:
> > @dleach02 As the description in commit message, this patch is only for zephyr. Thanks.
> 
> @yangbolu1991, this is not a good long term solution. We need to avoid maintaining SDK driver patches in the Zephyr HAL as it complicates updates of SDK drivers.

Is there documentation that explains where and how to add Upstream community Pull Requests to the SDK?

yangbolu1991:
> > @dleach02 As the description in commit message, this patch is only for zephyr. Thanks.
> 
> @yangbolu1991, this is not a good long term solution. We need to avoid maintaining SDK driver patches in the Zephyr HAL as it complicates updates of SDK drivers.

Understand. That should be a problem during SDK NETC driver initial design.
I have had more investigation in code, and resolved this in zephyr driver.
So, let me close the PR.

yangbolu1991:
> > > @dleach02 As the description in commit message, this patch is only for zephyr. Thanks.
> > 
> > 
> > @yangbolu1991, this is not a good long term solution. We need to avoid maintaining SDK driver patches in the Zephyr HAL as it complicates updates of SDK drivers.
> 
> Is there documentation that explains where and how to add Upstream community Pull Requests to the SDK?

I'd like to know the process too:)
Thanks.

PR #532, opened 2025-05-11
add imx943 Cortex-A Core support

JiafeiPan:
Added imx943 Cortex-A Core.


JiafeiPan:
rebased.

JiafeiPan:
Rebased.

decsny (approved):
since this is only adding files, merging it should be fine. but zephyr side still needs to be merged only after other outstanding hal manifest updates

PR #536, opened 2025-05-13
drivers: flash_k4: update to support MCXN series

DerekSnell:
Enables MCXN series to use this flash driver instead of ROM APIs.  Adds PFLASH feature macros to the MCXNx4x derivatives.


butok:
"It looks like MCXN23x is missing. Could it be added to cover the full MCXN family?

DerekSnell:
> It looks like MCXN23x is missing. Could it be added to cover the full MCXN family?

Hi @butok ,
Yes, the goal is to enable all the MCXN and MCXA devices that use this flash.  The features macros are generated for the devices, and still need to be generated for MCXN23x and some MCXA devices.  Once available, they can also be updated to use this driver.

PR #539, opened 2025-05-15
mcux: netc: ptp_clock: Correct function NETC_TimerAdjustFreq

ZhaoQiang-b45475:
The TMR_CTRL[period] is integral period of the timer clock in ns, while TMR_ADD represents the fractional clock period. So TMR_CTRL[period] should be updated too.


JiafeiPan:
Update copyright and also need to push to mcux-sdk-ng.

ZhaoQiang-b45475:
> Update copyright and also need to push to mcux-sdk-ng.

Thanks for your comment, fixed.

PR #540, opened 2025-05-15
devices: MIMX9596: fsl_memory: corrected the starting value of M7 memory offset

yongxu-wang15:
find in support edma test case, the start address is not right.


yongxu-wang15:
find in support edma test case, the start address is not right, need include this start value

decsny (dismissed):
you need to have a zephyr PR linked to this PR by the manifest change in order for us to be able to merge this.

mmahadevan108:
@JiafeiPan , can you help review

JiafeiPan:
Please also push this update to mcux-sdk NG.

PR #541, opened 2025-05-20
mcux: ieee802154: enabled any mode to mcxw ieee802154 driver

SynchronicIT:
Switching on IEEE 802.15.4 dependencies at any mode instead of only when using openthread or NET_L2.


mcux/middleware/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-if(CONFIG_BT OR CONFIG_NET_L2_IEEE802154 OR CONFIG_NET_L2_OPENTHREAD)
```
George-Stefan, line 1: `CONFIG_NET_L2_OPENTHREAD` can't be remove.
It's used for redfinch (RW) who has a RCP mode.

^^ @JA-NXP

George-Stefan (changes requested):

JarEmb (approved):

xavraz (approved):

George-Stefan (approved):

decsny (approved):

PR #542, opened 2025-05-20
mcux-sdk-ng: flexio: spi: Fix SPI master CS continuous issue

zejiang0jason:
- With current FLEXIO SPI master CS continuous mode implementation, the TX shifter register must be filled in time, otherwise the CS won't be continuous.
- Before CS continuous mode was supported, the function FLEXIO_SPI_MasterTransferNonBlocking:
  1. When it is called to transfer data, it fills one data frame to TX shifter buffer register first, then enables the FLEXIO interrupt.
  2. Only enables the RX shift buffer register full interrupt, reads RX data and fills TX data in ISR.
- This flow can't meet CS continuous mode requirement, because: After the first data frame written to TX shifter buffer register, the second data frame is only written after RX interrupt asserts. During this period, the TX shifter is not filled in time. CS is not continuous.
- To fix this issue, the TX interrupt is also enabled in CS continuous mode, to make the second TX data be filled in time in ISR. But this fix is not good:
  - If FLEXIO interrupt is blocked by other higher priority interrupt, and not served in time, the CS will still not be continuous.
  - In the ISR, when no interrupt status flag assert (https://github.com/zephyrproject-rtos/hal_nxp/pull/221), the ISR should return directly, but current fix assumes TX interrupt occurs and writes TX shifter buffer.
- This patch uses a new way to implement the CS continuous mode:
  - Only enable RX interrupt, don't enable TX interrupt
  - Change the timer1 working mode, to make sure CS is always low during whole data transfer, it is not necessary to always fill TX shifter buffer in time.

### DIFF
--- a/mcux/mcux-sdk-ng/drivers/flexio/spi/fsl_flexio_spi.c
+++ b/mcux/mcux-sdk-ng/drivers/flexio/spi/fsl_flexio_spi.c
@@ -52,6 +52,41 @@ static void FLEXIO_SPI_TransferSendTransaction(FLEXIO_SPI_Type *base, flexio_spi
  */
 static void FLEXIO_SPI_TransferReceiveTransaction(FLEXIO_SPI_Type *base, flexio_spi_master_handle_t *handle);
 
+/*!
+ * @brief Set the Timer 1 register TIMCFG
+ *
+ * This function is used for master mode. It sets the Timer 1 (used to control CS pin)
+ * TIMCFG register based on CS continous mode configuration, and return the old
+ * register value, for later recovery.
+ *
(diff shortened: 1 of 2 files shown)


mmahadevan108:
@decsny , please revisit.

decsny:
Are these changes also going in SDK ?

zejiang0jason:
Hi @decsny ,yes, the patch has been merged in SDK.

dleach02:
@mmahadevan108 Zephyr side is ready and this patch has already been merged into the SDK.

dleach02 (approved):

PR #554, opened 2025-06-09
mcux-sdk-ng: enet: fsl_enet: avoid possible race condition on timestamp

theadib:
Only correct second value if the nanosecond has reasonably value.

If the overflow happen short after the capture of the timestamp,
but before checking the ISR flag,
then the second value is corrected falsely.

The patch applies for the FRDM_K64F board having K64 family.


theadib:
Hello @mmahadevan108  I have done the same PR to the mcux-sdk
https://github.com/nxp-mcuxpresso/mcux-sdk/pull/236
regards Adib

mmahadevan108:
Once merged into the SDK repo then we can take this change here as we do not have to worry about losing this fix during SDK updates in the HAL.

PR #559, opened 2025-06-13
mcux: wifi_bt_module: update IW610 TX power table

MaochenWang1:
Update IW610 M2 TX power limit table.


dleach02:
Can we plan on moving these tables to the boards folder? It is really problematic to have this in the HAL.

PR #569, opened 2025-06-26
Added some expected defines that were required by soc.c in zephyr

imran-sajjad:
ZEPHYR/zephyr/soc/nxp/imxrt/imxrt10xx/soc.c
expects kCLOCK_Sai2Mux, kCLOCK_Sai2PreDiv, kCLOCK_Sai2Div to be defined but these were not defined for the mimxrt1011 family of devices. This fixes the compile error when doing anything with i2s with the 1010 family of devices.

Other possible fixes could be to modify soc.c to have cases for every board or to have different soc.c for every board. Also kCLOCK_Sai2Mux could be set to a known, invalid value in fsl_clock.h to cause a compile time error if being used.

### DIFF
--- a/mcux/mcux-sdk/devices/MIMXRT1011/drivers/fsl_clock.h
+++ b/mcux/mcux-sdk/devices/MIMXRT1011/drivers/fsl_clock.h
@@ -510,6 +510,10 @@ typedef enum _clock_mux
                                CCM_CSCMR1_SAI3_CLK_SEL_SHIFT,
                                CCM_CSCMR1_SAI3_CLK_SEL_MASK,
                                CCM_NO_BUSY_WAIT), /*!< sai3 mux name */
+    kCLOCK_Sai2Mux       = CCM_TUPLE(CSCMR1_OFFSET,
+                               CCM_CSCMR1_SAI1_CLK_SEL_SHIFT,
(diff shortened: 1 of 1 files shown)


ZhaoxiangJin:
Seems this [commit](https://github.com/zephyrproject-rtos/hal_nxp/pull/569/commits/3c8863aa27c833c484b684c040681df2b1d0804b) is not necessary? Add @lucien-nxp to review.

imran-sajjad:
Thank you for responding so fast. Yes I agree it's not the prettiest fix but at least you are aware of the issue now.

I guess there is an argument to be made that if the soc.c file is "meant" to be device agnostic, then modifying it to be board aware defeats that purpose. Whether that's more important rather than making the interface depend on the available hardware is the author's call.

It looks like these enums are being used like a public interface. Would it be possible to set those enums values for sai2 to known invalid values that the soc.c file can be aware of? Like

kCLOCK_Sai3Mux = CCM_TUPLE(...)
kCLOCK_Sai2Mux = CCM_INVALID
kCLOCK_Sai1Mux = CCM_TUPLE(...)

imran-sajjad:
Nevermind, it looks like there is already an ifdef for a board in the soc.c file. We can close this PR and I can raise an issue in the zephyr repo.

Thanks for looking at it

PR #570, opened 2025-06-27
dts: nxp: mcx: Updated Pinctrl Mask for MCXW device.

EmilioCBen:
Updated pinctrl mask on MCXW device from 0x7 -> 0xF Register is a range of 4 bit values.


dleach02:
@EmilioCBen please investigate the tooling that generated this error. I want to understand how or why this occurred.

EmilioCBen:
> @EmilioCBen please investigate the tooling that generated this error. I want to understand how or why this occurred.

I have found the issue in the tooling, I will investigate a fix to the issue and create a PR.

PR #571, opened 2025-07-03
enable NETC on imx95 Cortex-A55

JiafeiPan:
1. drivers: netc: make the driver to be aarch64 compatible
2. mcux-sdk: MIMX9596_ca55: disable FSL_FEATURE_NETC_HAS_ERRATA_051711
3. mcux-sdk: MIMX9596_ca55: fix A-Core header file regard of netc struct definition
4. fsl_common_arm: Fix comment sign issue on different architectures

### DIFF
--- a/mcux/mcux-sdk-ng/drivers/common/fsl_common_arm.h
+++ b/mcux/mcux-sdk-ng/drivers/common/fsl_common_arm.h
@@ -474,15 +474,21 @@ _Pragma("diag_suppress=Pm120")
 #endif
 
 #elif (defined(__GNUC__)) || defined(DOXYGEN_OUTPUT)
+#if defined(__ARM_ARCH_8A__) /* This macro is ARMv8-A specific */
+#define MCUX_CS "//"
+#else
+#define MCUX_CS "@"
+#endif
+
 /* For GCC, when the non-cacheable section is required, please define "__STARTUP_INITIALIZE_NONCACHEDATA"
  * in your projects to make sure the non-cacheable section variables will be initialized in system startup.
  */
 #define AT_NONCACHEABLE_SECTION_INIT(var) __attribute__((section("NonCacheable.init"))) var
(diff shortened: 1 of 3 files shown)


mmahadevan108:
@JiafeiPan , have the changes to the SDK driver been submitted to the SDK team?

mcux/mcux-sdk-ng/drivers/common/fsl_common_arm.h
```diff
@@ -474,15 +474,21 @@ _Pragma("diag_suppress=Pm120")
 #endif
 
 #elif (defined(__GNUC__)) || defined(DOXYGEN_OUTPUT)
+#if defined(__ARM_ARCH_8A__) /* This macro is ARMv8-A specific */
+#define __CS "//"
+#else
+#define __CS "@"
```
zejiang0jason, line 480: Hello, because of MISRA C 2012 rule 21.1: https://stackoverflow.com/questions/50696794/misra-c-2012-rule-21-1-macros-starting-with-underscore. We will not use macro name started with underscore.
Also, this file will be included by every SDK file, to make it not conflict with other code, suggest adding prefix, like `MCUX_CS`. BTW, what is the meaning of CS? thanks.

yangbolu1991:
Except Jason's comments, I am ok with other changes.
Will need @JiafeiPan to send the changes to SDK too.

Thanks.

mcux/mcux-sdk-ng/drivers/common/fsl_common_arm.h
JiafeiPan: CS is comment sign character, so let's change it to be "MCUX_CS", thanks.

JiafeiPan:
The "mcux-sdk: MIMX9596_ca55: fix A-Core header file" is only for Cortex-A core and no need to push it to mcux-sdk NG, the other three patches have been merged into SDK.

zejiang0jason (approved):

JiafeiPan:
rebased.

PR #580, opened 2025-07-22
mcux: mcux-sdk: devices: fix MKV56F24.h wrong macro

nxp-wei-feng:
The MKV56F24 device doesn't contain CAN2 peripheral. Remove it from the device header.


mmahadevan108:
Has this fix been included in the SDK repo?

dleach02:
@nxp-wei-feng please add a comment in the mcux/README highlighting the patch. We probably won't be doing any broad updates of the mcux-sdk directory but just in case.

nxp-wei-feng:
> Has this fix been included in the SDK repo?

@mmahadevan108 this is a very old device which is not supported in SDK next. So we need to update mcux-sdk directly.

PR #597, opened 2025-09-02
drivers: xspi: cache64: Fix cache management issue

SuperHeroAbner:
Fix issue that the calculation of address is wrong. The last byte is not handled by cache api.


zejiang0jason:
@Albort12138 to review the change.

ZhaoxiangJin:
Hello @SuperHeroAbner, is this patch unique to hal_nxp? Is this patch also in the SDK?

SuperHeroAbner:
@ZhaoxiangJin If it's a real issue. It should be also in SDK. Currently just fix this Zephyr test problem.

dleach02:
Please coordinate this patch with the SDK-NG team and their update

mcux/README
```diff
@@ -239,3 +239,6 @@ Patch List:
    - This patch is in hal_nxp but is not in mcux-sdk-ng 25.09.00.
      This patch was merged after PR https://github.com/zephyrproject-rtos/hal_nxp/pull/612/commits was created, so it
      did not enter PR#612 via cherry-pick but rather via rebasing onto the master branch.
+
+  - mcux-sdk-ng: drivers: xspi: cache64: Fix cache management issue.
+   - This patch is in hal_nxp but is not in mcux-sdk-ng 25.09.00.
```
SuperHeroAbner: @dleach02 Done. Is it as expected? We still haven't put it in 25.12.00 release. I will update SDK related code to catch the 25.12.00.

mmahadevan108:
@ZhaoxiangJin @SuperHeroAbner has this merged into the SDK repo?

PR #603, opened 2025-09-15
Fix rt1180 pinctrl error

lucien-nxp:
Regenerate pinctrl files for RT1180 using modified script:
1. Ignore some bit field value is a range
2. When a peripheral name contains a bracketed index like
  [x] at the end, it should be transformed into _x.
3. Special cases on RT1180 xbara:
  need to handle configuration id condition because
  the content will be scanned by script for xbara signal


mcux/scripts/pinctrl/gen_soc_headers.py
```diff
@@ -129,7 +129,7 @@ def main():
 
     data_version = get_pack_version(temp_dir.name)
     print(f"Found data pack version {data_version}")
-    if round(data_version) != 16:
+    if round(data_version) != '25.06':
         print("Warning: This tool is only verified for data pack version 16, "
```
danieldegrasse, line 133: May want to update the version error message here as well- I also typically would touch the readme files when updating the script in the past

lucien-nxp: Sorry, I forget to modify it. Modified.

PR #704, opened 2026-04-01
mcux: utilities: Add optimized assembly memset implementation

zejiang0jason:
- Add fsl_memset.S, an ARM Thumb assembly memset that replaces the slow byte-by-byte newlib nano implementation with a word-aligned bulk-fill using stmia (16 bytes/loop) for Cortex-M cores
 - Handle unaligned leading bytes, then fill 16/8/4/2/1-byte remainders


zejiang0jason:
this has been submited to sdk, but not merged yet.

fengming-ye (approved):

MaochenWang1 (approved):

ZhaoxiangJin (approved):

PR #705, opened 2026-04-01
enet: take coherent snapshots of RX descriptors

hakehuang:
Fix a race in the NXP ENET HAL RX path where software reads RX buffer descriptor fields directly while hardware/DMA may still be updating them.

This change:
- adds a helper to take a coherent RX descriptor snapshot
- uses that snapshot in `ENET_GetRxFrameSize()` and `ENET_ReadFrame()`
- adds `__DMB()` ordering barriers when returning RX descriptors to hardware in `ENET_UpdateReadBuffers()`

Without this, newer toolchains can expose torn RX descriptor reads, which can cause received frames to be missed or treated as invalid. On Zephyr this shows up as NXP ENET DHCP staying in `selecting` even though link is up.

Related:
- zephyrproject-rtos/zephyr#106150

Validation:
- code inspection confirms Zephyr `eth_nxp_enet_rx()` depends on `ENET_GetRxFrameSize()` / `ENET_ReadFrame()`
- remote test environment reproducer uses `samples/net/dhcpv4_client` on `mimxrt1170_evk@B/mimxrt1176/cm7`
- remote environment shows Zephyr SDK 1.0.0 builds are affected while a Zephyr SDK 0.17.4 build is known-good in the same setup
- the unfixed remote hal_nxp revision does not contain the RX snapshot helper, while this branch adds it

A companion Zephyr PR will update `west.yml` to this hal_nxp revision.

### DIFF
--- a/mcux/mcux-sdk-ng/drivers/enet/fsl_enet.c
+++ b/mcux/mcux-sdk-ng/drivers/enet/fsl_enet.c
@@ -155,13 +155,56 @@ static status_t ENET_SetRxBufferDescriptors(ENET_Type *base,
                                             const enet_buffer_config_t *bufferConfig);
 
 /*!
+ * @brief Releases one ENET read buffer descriptor back to hardware.
+ *
+ * @param handle The ENET handle pointer.
+ * @param ringId The descriptor ring index, range from 0 ~
+ *        (FSL_FEATURE_ENET_INSTANCE_QUEUEn(x) - 1).
+ */
+static void ENET_ReleaseRxBufferDescriptor(enet_handle_t *handle, uint8_t ringId);
+
+/*!
  * @brief Updates the ENET read buffer descriptors.
  *
  * @param base ENET peripheral base address.
  * @param handle The ENET handle pointer.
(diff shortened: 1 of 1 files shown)


stanislav-poboril (changes requested):
Similar changes might be needed in ENET_GetRxErrBeforeReadFrame, ENET_GetRxFrame and ENET_GetRxFrameErr (as it is called from ENET_GetRxFrameErr only, we can pass the snapshot there). Only ENET_GetRxErrBeforeReadFrame is used from Zephyr, the other changes could be made in MCUXpresso SDK (which will then propagate to HAL later).

mmahadevan108:
@hakehuang , can you please revisit this PR based on feedback.

stanislav-poboril (approved):

mmahadevan108:
@stanislav-poboril can you help take this change back to the SDK driver repo

stanislav-poboril:
> @stanislav-poboril can you help take this change back to the SDK driver repo

@mmahadevan108 sure, will put it in SDK 26.06.00.

mmahadevan108 (approved):

mmahadevan108:
@hakehuang , do we need anything on the Zephyr side to integrate this HAL change

PR #709, opened 2026-04-09
wifi_nxp: wlcmgr: Fix missing 'WLAN_REASON_UAP_SUCCESS' event and dec…

mfumita:
…ouple from L3 state

Applications waiting for the 'WIFI_STATUS_AP_SUCCESS' event from the NXP Wi-Fi driver (zephyr/drivers/wifi/nxp/nxp_wifi_drv.c) would either never receive it or experience inconsistent behavior depending on the 'CONFIG_WIFI_NM_WPA_SUPPLICANT' setting.

Prior to this commit, this was caused by two issues:
1. The 'WLAN_REASON_UAP_SUCCESS' event trigger condition arbitrarily switched between waiting for 'CM_UAP_STARTED' (L2) and 'CM_UAP_IP_UP' (L3). When 'CONFIG_WIFI_NM_WPA_SUPPLICANT' was set, the event was triggered while 'case WIFI_EVENT_UAP_STARTED:' was processed in 'uap_state_machine()'. However, when 'CONFIG_WIFI_NM_WPA_SUPPLICANT' was NOT set, it was triggered while 'case WIFI_EVENT_UAP_NET_ADDR_CONFIG:' was processed. Therefore, in L2-only uAP scenarios without 'CONFIG_WIFI_NM_WPA_SUPPLICANT', it was never triggered because no IP address is set.
2. The 'WLAN_REASON_UAP_SUCCESS' event was triggered inside 'uap_state_machine()' before the global 'wlan.uap_state' was updated. This caused the state check inside 'wlan_get_current_uap_network_ssid()' to fail, dropping the event from 'nxp_wifi_drv.c' => 'nxp_wifi_wlan_event_callback()'.

This commit fixes these issues by removing the event triggers from 'uap_state_machine()'. Instead, it unifies the trigger condition to simply wait for the 'CM_UAP_STARTED' state, ensuring consistent behavior and properly supporting IP-less setups.
It also moves the event trigger to 'wlcmgr_task()' after 'wlan.uap_state' is updated.

### DIFF
--- a/mcux/middleware/wifi_nxp/wlcmgr/wlan.c
+++ b/mcux/middleware/wifi_nxp/wlcmgr/wlan.c
@@ -6164,7 +6164,6 @@ static enum cm_uap_state uap_state_machine(struct wifi_message *msg)
                 t_u8 sta_channel = mlan_adap->priv[0]->curr_bss_params.bss_descriptor.channel;
                 t_u8 uap_channel = mlan_adap->priv[1]->uap_channel;
 
-                CONNECTION_EVENT(WLAN_REASON_UAP_SUCCESS, NULL);
                 if(is_sta_connected() && uap_channel != sta_channel)
                 {
                     while(!is_uap_started())
@@ -6300,7 +6299,6 @@ static enum cm_uap_state uap_state_machine(struct wifi_message *msg)
                 (void)net_get_if_ipv6_addr((struct net_ip_config *)&network->ip, if_handle);
 #endif
                 next = CM_UAP_IP_UP;
-                CONNECTION_EVENT(WLAN_REASON_UAP_SUCCESS, NULL);
             }
             else
             {
@@ -7574,6 +7572,9 @@ static void wlcmgr_task(void *data)
 
                 wlcm_d("SM uAP %s -> %s", dbg_uap_state_name(wlan.uap_state), dbg_uap_state_name(next_uap_state));
                 wlan.uap_state = next_uap_state;
+                if (wlan.uap_state == CM_UAP_STARTED) {
+                    CONNECTION_EVENT(WLAN_REASON_UAP_SUCCESS, NULL);
+                }
 #else
                 wlcm_w("UAP feature disabled recv wlcm msg %d", msg.event);
 #endif


MaochenWang1:
> 1. Therefore, in L2-only uAP scenarios without 'CONFIG_WIFI_NM_WPA_SUPPLICANT', it was never triggered because no IP address is set.

How to configure this `L2-only uAP`?

MaochenWang1:
> 2\. The 'WLAN_REASON_UAP_SUCCESS' event was triggered inside 'uap_state_machine()' before the global 'wlan.uap_state' was updated. This caused the state check inside 'wlan_get_current_uap_network_ssid()' to fail, dropping the event from 'nxp_wifi_drv.c' => 'nxp_wifi_wlan_event_callback()'.

This is the `CONFIG_WIFI_NM_WPA_SUPPLICANT` **NOT defined** case, right?
`WLAN_REASON_UAP_SUCCESS` is triggered in `case WIFI_EVENT_UAP_NET_ADDR_CONFIG`, but current `cm_uap_state` should be `CM_UAP_STARTED`, why there is check error in `wlan_get_current_uap_network_ssid()`?

mfumita:
> > 1. Therefore, in L2-only uAP scenarios without 'CONFIG_WIFI_NM_WPA_SUPPLICANT', it was never triggered because no IP address is set.
> 
> How to configure this `L2-only uAP`?

Initially, we do not want to assign an IP address to the Wi-Fi interface in Zephyr at all. We just start the AP and expect it to operate purely at L2.

We are building an Ethernet-to-Wi-Fi AP bridge using the u-blox IRIS-W1 module. The architecture involves a Host machine connected via Ethernet. This Host machine acts as the master and handles all L3 configurations (like DHCP and IP assignment) for the bridged network.

Therefore, on the Zephyr side, we want the u-blox module to operate solely as an L2 bridge, and we do not need to configure any IP address for the Wi-Fi interface. To add the Wi-Fi interface to the bridge, our Zephyr application waits for the NXP Wi-Fi driver to emit the 'WIFI_STATUS_AP_SUCCESS' event indicating the AP is up.

However, in the current driver implementation, if we don't configure an IP, the L3 state ('CM_UAP_IP_UP') is never reached, and the AP success event is never triggered.

As a side note, Zephyr's Ethernet bridge implementation does not natively support Wi-Fi interfaces yet, which I am addressing in a separate PR (zephyrproject-rtos/zephyr#107086). But this L2-only AP startup fix in the NXP driver is a prerequisite for that bridging architecture to work.

mfumita:
> > 2. The 'WLAN_REASON_UAP_SUCCESS' event was triggered inside 'uap_state_machine()' before the global 'wlan.uap_state' was updated. This caused the state check inside 'wlan_get_current_uap_network_ssid()' to fail, dropping the event from 'nxp_wifi_drv.c' => 'nxp_wifi_wlan_event_callback()'.
> 
> This is the `CONFIG_WIFI_NM_WPA_SUPPLICANT` **NOT defined** case, right? `WLAN_REASON_UAP_SUCCESS` is triggered in `case WIFI_EVENT_UAP_NET_ADDR_CONFIG`, but current `cm_uap_state` should be `CM_UAP_STARTED`, why there is check error in `wlan_get_current_uap_network_ssid()`?

You are correct that if an IP address is assigned, the event triggers at 'WIFI_EVENT_UAP_NET_ADDR_CONFIG' and the check in 'wlan_get_current_uap_network_ssid()' passes.

However, as I mentioned in the earlier comment, we do not assign an IP address, so the state machine never reaches 'WIFI_EVENT_UAP_NET_ADDR_CONFIG'.

To fix this no-IP scenario, we need to trigger the success event during 'WIFI_EVENT_UAP_STARTED'. But if we put the event call inside that switch case, we hit the problem described in Issue 2: the event fires while we are still inside 'uap_state_machine()'. At that moment, the global 'wlan.uap_state' has not been updated yet. Because the state is outdated, 'wlan_get_current_uap_network_ssid()' fails and drops the event.

MaochenWang1:
> we do not want to assign an IP address to the Wi-Fi interface in Zephyr at all. We just start the AP and expect it to operate purely at L2

What modifications are needed to achieve this? I want to reproduce this issue locally.

mfumita:
> > we do not want to assign an IP address to the Wi-Fi interface in Zephyr at all. We just start the AP and expect it to operate purely at L2
> 
> What modifications are needed to achieve this? I want to reproduce this issue locally.

I have prepared a minimal sample logic to start a Wi-Fi AP without an IP address below.
This is for the ubx_evk_iris_w1 board.

Firstly, since the current NXP Wi-Fi driver requires a DHCP server on the Wi-Fi AP,
it needs to be disabled by cherry-picking the following PR.
https://github.com/zephyrproject-rtos/zephyr/pull/107171

[applications\test\boards\ubx_evk_iris_w1.conf]
```
CONFIG_NXP_RW610=y
CONFIG_NXP_WIFI_SOFTAP_SUPPORT=y

CONFIG_WIFI_NXP=y
CONFIG_WIFI_NM_MAX_MANAGED_INTERFACES=2
CONFIG_NXP_WIFI_SOFTAP_DHCP_SERVER=n
```


[applications\test\prj.conf]
```
CONFIG_MAIN_STACK_SIZE=2048

CONFIG_NETWORKING=y
CONFIG_WIFI=y
CONFIG_NET_IPV4=y
CONFIG_NET_IPV6=n

CONFIG_LOG=y

CONFIG_SHELL=y
CONFIG_NET_SHELL=y
```


[applications\test\main.c]
```
#include <zephyr/kernel.h>
#include <zephyr/logging/log.h>
#include <zephyr/net/wifi_mgmt.h>
#include <zephyr/net/net_if.h>

LOG_MODULE_REGISTER(MAIN);

static const char* tst_ssid = "test";

static struct net_if *ap_iface;
static struct wifi_connect_req_params ap_config;
static struct net_mgmt_event_callback cb;

static void wifi_event_handler(struct net_mgmt_event_callback *cb, uint64_t mgmt_event,
			       struct net_if *iface)
{
	LOG_INF("AP Mode is enabled.");
}

static void enable_ap_mode(void)
{
	if (!ap_iface) {
		LOG_INF("AP: is not initialized");
		return;
	}

	LOG_INF("Turning on AP Mode");
	ap_config.ssid = (const uint8_t *)tst_ssid;
	ap_config.ssid_length = sizeof(tst_ssid) - 1;
	ap_config.channel = WIFI_CHANNEL_ANY;
	ap_config.band = WIFI_FREQ_BAND_2_4_GHZ;

	int ret = net_mgmt(NET_REQUEST_WIFI_AP_ENABLE, ap_iface, &ap_config,
			   sizeof(struct wifi_connect_req_params));
	if (ret) {
		LOG_ERR("NET_REQUEST_WIFI_AP_ENABLE failed, err: %d", ret);
	}
}

int main(void)
{
	net_mgmt_init_event_callback(&cb, wifi_event_handler, NET_EVENT_WIFI_AP_ENABLE_RESULT);
	net_mgmt_add_event_callback(&cb);

	ap_iface = net_if_get_wifi_sap();
	enable_ap_mode();

	return 0;
}
```


[applications\test\CMakeLists.txt]
```
cmake_minimum_required(VERSION 3.20.0)

find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
project(test)

target_sources(app PRIVATE src/main.c)
```

Then run build.
```
west build -b ubx_evk_iris_w1 applications/test/
```

You can confirm that an IP address isn't assigned to the Wi-Fi AP interface by running 'net ipv4' command in the shell.

In this program, the message "AP Mode is enabled." does not appear because the 'NET_EVENT_WIFI_AP_ENABLE_RESULT' event is never triggered.
However, once the patch in this PR is applied, the event is triggered.

mfumita:
> > > 2. The 'WLAN_REASON_UAP_SUCCESS' event was triggered inside 'uap_state_machine()' before the global 'wlan.uap_state' was updated. This caused the state check inside 'wlan_get_current_uap_network_ssid()' to fail, dropping the event from 'nxp_wifi_drv.c' => 'nxp_wifi_wlan_event_callback()'.
> > 
> > 
> > This is the `CONFIG_WIFI_NM_WPA_SUPPLICANT` **NOT defined** case, right? `WLAN_REASON_UAP_SUCCESS` is triggered in `case WIFI_EVENT_UAP_NET_ADDR_CONFIG`, but current `cm_uap_state` should be `CM_UAP_STARTED`, why there is check error in `wlan_get_current_uap_network_ssid()`?
> 
> You are correct that if an IP address is assigned, the event triggers at 'WIFI_EVENT_UAP_NET_ADDR_CONFIG' and the check in 'wlan_get_current_uap_network_ssid()' passes.
> 

I mentioned that the event is called when an IP is assigned, but it seems to happen only for IPv6.
So the event still doesn't seem to get triggered even when assigning an IPv4 address.

'WIFI_EVENT_UAP_NET_ADDR_CONFIG' seems to be triggered only from the following place.
[modules\hal\nxp\mcux\middleware\wifi_nxp\port\net\zephyr\net.c]
wifi_net_event_handler() => switch (mgmt_event) => case NET_EVENT_IPV6_DAD_SUCCEED:

MaochenWang1:
Hi @mfumita  could you create a zephyr github issue about this issue? I want to find a proper way to fix your issue.

MaochenWang1:
Hi @mfumita  can you give a try if my proposal https://github.com/zephyrproject-rtos/zephyr/pull/112346 can fix your issue?
Please let me know the result, then we can merge it.

mfumita:
> Hi @mfumita can you give a try if my proposal [zephyrproject-rtos/zephyr#112346](https://github.com/zephyrproject-rtos/zephyr/pull/112346) can fix your issue? Please let me know the result, then we can merge it.

While it might fix the issue of sending 'WIFI_EVENT_UAP_NET_ADDR_CONFIG' when IPv6 is disabled
(I haven't verified this specific case), this is a different issue from my original report. 

My original problem is that the AP mode doesn't work in an L3-less (pure L2) configuration.
Since we don't assign an IP address at all, the AP still doesn't work even with this patch
because that event is never triggered.

Therefore, this doesn't resolve the issue for my use case.

MaochenWang1:
> While it might fix the issue of sending 'WIFI_EVENT_UAP_NET_ADDR_CONFIG' when IPv6 is disabled (I haven't verified this specific case), this is a different issue from my original report.
> 
> My original problem is that the AP mode doesn't work in an L3-less (pure L2) configuration. Since we don't assign an IP address at all, the AP still doesn't work even with this patch because that event is never triggered.
> 

Hi @mfumita I used the conf files and example code you provided from this conversation https://github.com/zephyrproject-rtos/hal_nxp/pull/709#issuecomment-4256962550 to test (I thought it could reproduce your original issue), and the fix in https://github.com/zephyrproject-rtos/zephyr/pull/112346 can work fine and fix the issue.
What is L3-less (pure L2) configuration for AP mode? how to achieve this? Maybe you can provide the configurations then I can try to reproduce the issue and find a proper fix.

mfumita:
> > While it might fix the issue of sending 'WIFI_EVENT_UAP_NET_ADDR_CONFIG' when IPv6 is disabled (I haven't verified this specific case), this is a different issue from my original report.
> > My original problem is that the AP mode doesn't work in an L3-less (pure L2) configuration. Since we don't assign an IP address at all, the AP still doesn't work even with this patch because that event is never triggered.
> 
> Hi @mfumita I used the conf files and example code you provided from this conversation [#709 (comment)](https://github.com/zephyrproject-rtos/hal_nxp/pull/709#issuecomment-4256962550) to test (I thought it could reproduce your original issue), and the fix in [zephyrproject-rtos/zephyr#112346](https://github.com/zephyrproject-rtos/zephyr/pull/112346) can work fine and fix the issue. What is L3-less (pure L2) configuration for AP mode? how to achieve this? Maybe you can provide the configurations then I can try to reproduce the issue and find a proper fix.

Actually, I confirmed that this solves the issue I reported.
I'm really sorry, but I was accidentally running an older version of the firmware before.
Thank you very much for your help.

PR #712, opened 2026-04-16
dts: nxp: rt104x: update and add pinctrl data

lucien-nxp:
- Update RT1041/RT1042 pinctrl dtsi files with corrected daisy register values and copyright year. (RT104X Data is generated from NXP Config Tool 26.03.)
- Add new pinctrl data for RT1043 and RT1046 SoC variants.

PR #713, opened 2026-04-17
Update rt pinctrl generation script

lucien-nxp:
- Integrate imx_fixup_pinmux into gen_soc_headers.py so RT 4-digit NPI updates cannot skip the daisy register fixup step. Require `--iomuxc-file` when IOMUX controller is used; the script errors out if missing rather than silently generating incorrect pinctrl files. Refactor imx_fixup_pinmux.py into reusable functions while preserving its CLI.
- Correct misspelling of "defintions" to "definitions" in docstrings, comments, help text, and generated file headers across the pinctrl generation scripts.

### DIFF
--- a/mcux/scripts/pinctrl/gen_soc_headers.py
+++ b/mcux/scripts/pinctrl/gen_soc_headers.py
@@ -20,6 +20,7 @@ import re
 
 # SOC configuration data support libraries
 from imx import imx_cfg_utils
+from imx import imx_fixup_pinmux
 from kinetis import kinetis_cfg_utils
 from lpc import lpc_cfg_utils
 
@@ -50,6 +51,9 @@ def parse_args():
     parser.add_argument('--controller', metavar = 'CTRL', type=str,
                         help=("SOC pin controller type."
                         "Currently supports: [IOMUX, IOCON, PORT]"))
+    parser.add_argument('--iomuxc-file', metavar = 'IOMUXC', type=str,
+                        help=('Path to fsl_iomuxc.h file for daisy register fixup. '
+                        'Required for IOMUX controller (i.MX RT 4-digit parts only)'))
 
     return parser.parse_args()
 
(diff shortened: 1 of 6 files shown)


lucien-nxp:
Hi @mmahadevan108 @hakehuang @zejiang0jason @Holt-Sun @ZhaoxiangJin 
Could you help review the PR for pinctrl generated script update? This PR don't need to be tested on zephyr side because of only script update where there is no files used by zephyr side.
I found that the update of the `daisy `register in `iomuxc `heavily relies on re-checking after generating the initialization version by `imx_fixup_pinmux.py`. This is very easy to overlook for new developers. Therefore, I integrated it into the general generation script to avoid such omissions.

lucien-nxp:
RT104X pinctrl files have been generated by updated script. #712

hakehuang:
@lucien-nxp can you add unity testing for such scripts?

lucien-nxp:
> @lucien-nxp can you add unity testing for such scripts?

Sorry, I don't want to add unity test because I think unit tests can only verify that the extraction logic runs without error, not that the extracted pin definitions are electrically correct. The actual correctness depends on the upstream data pack, which can only be validated at the hardware level.

hakehuang:
> Sorry, I don't want to add unity test because I think unit tests can only verify that the extraction logic runs without error, not that the extracted pin definitions are electrically correct. The actual correctness depends on the upstream data pack, which can only be validated at the hardware level.

I do a regression on mimxrt1040_evk with v4.4.0-527-g801122041677, which has no regression issues. but as I said, HW testing can not cover all the PINs and its potential alternatives. A unit testing can help to make sure the converting is precise no data missing or wrong converting, it is very valuable work. But it is not a blocker for this pr

lucien-nxp:
> > Sorry, I don't want to add unity test because I think unit tests can only verify that the extraction logic runs without error, not that the extracted pin definitions are electrically correct. The actual correctness depends on the upstream data pack, which can only be validated at the hardware level.
> 
> I do a regression on mimxrt1040_evk with v4.4.0-527-g801122041677, which has no regression issues. but as I said, HW testing can not cover all the PINs and its potential alternatives. A unit testing can help to make sure the converting is precise no data missing or wrong converting, it is very valuable work. But it is not a blocker for this pr

Agreed. Adding unit tests is a valuable follow-up, but I'd treat it as a separate topic — to be useful, the tests would need to cover all NXP SoC series, which makes it a fairly large effort on its own. Since it's not urgent at the moment, I'll file a ticket and revisit the script when I have bandwidth.

lucien-nxp:
Hi @zejiang0jason @ZhaoxiangJin 
Could you help review and merge this PR firstly?

ZhaoxiangJin (approved):

PR #716, opened 2026-04-21
zephyr: support compressed or uncompressed FW for RW612 and IW610

MaochenWang1:
Use a shared suffix to select compressed or uncompressed Wi-Fi and NBU firmware binaries based on CONFIG_NXP_COMPRESSED_WIFI_NB_FW. Using a compressed firmware reduces flash memory consumption. However, the firmware must be decompressed during initialization after being downloaded, which increases boot time and initialization latency.

### DIFF
--- a/zephyr/module.yml
+++ b/zephyr/module.yml
@@ -30,29 +30,53 @@ blobs:
 
 #For RW61x
   - path: rw61x/rw61x_sb_ble_15d4_combo_a2.bin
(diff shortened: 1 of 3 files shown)


axelnxp (approved):

MaochenWang1:
add a new commit: update module.yml to fetch RW612 and IW610 FW

mmahadevan108 (approved):

PR #720, opened 2026-05-04
mcux: mimx8ml8: clock: Add missing media display clock root definitions

pwattebledsfl:
# Summary

Add two missing `clock_root_control_t` entries for MIMX8ML8 (i.MX8MP):

- `kCLOCK_RootMediaMipiPhy1Ref` — ROOT[123], MIPI DPHY PLL reference clock
- `kCLOCK_RootMediaDisp1Pix` — ROOT[124], LCDIF1 pixel clock

# Justification

These clock roots are listed in the i.MX8MP Reference Manual (IMX8MPRM) Table 5-1 "Clock Root Table" but are absent from the SDK enum. Any display pipeline using LCDIF1 + MIPI DSI on this SoC requires these roots. Currently, users must define them manually.

# Change

Two enum entries added after `kCLOCK_RootMediaApb`, using the same encoding pattern as all existing entries. No functional change to existing code.


RuoshanShi:
I have verified the i.MX8MP Reference Manual, and the CLOCK ROOT IDs used in the patch are correct. The patch has already been backported to the NXP SDK.

ZhaoxiangJin (approved):
Thanks for your contribution!

PR #723, opened 2026-05-09
Fix event configuration for Counter_H

einergehtnochrein:
HEVENT field must be set for all Counter_H events, not only match events. The User Manual is misleading here.

### DIFF
--- a/mcux/mcux-sdk-ng/devices/i.MX/i.MX952/MIMX9529/drivers/fsl_clock.h
+++ b/mcux/mcux-sdk-ng/devices/i.MX/i.MX952/MIMX9529/drivers/fsl_clock.h
@@ -8,9 +8,11 @@
 #ifndef _FSL_CLOCK_H_
 #define _FSL_CLOCK_H_
 #include "fsl_common.h"
+#ifndef CONFIG_ARM_SCMI
 #include "scmi.h"
 #include "scmi_clock.h"
 #include "scmi_internal.h"
+#endif
 
 /*!
  * @brief CCM reg macros to map corresponding registers.
--- a/mcux/mcux-sdk-ng/devices/i.MX/i.MX952/MIMX9529/system_MIMX9529_cm7.h
+++ b/mcux/mcux-sdk-ng/devices/i.MX/i.MX952/MIMX9529/system_MIMX9529_cm7.h
@@ -53,9 +53,11 @@ extern "C" {
 #endif
 
 #include <stdint.h>
+#ifndef CONFIG_ARM_SCMI
 #include "scmi.h"
 #include "scmi_internal.h"
 #include "smt.h"
+#endif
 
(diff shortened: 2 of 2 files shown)


ZhaoxiangJin:
Hello @FelixWang47831, please help review, thanks.

einergehtnochrein:
I just realized I left out the context. This PR is about the SCTimer component in mcux-sdk-ng, and applies to all SCTimer instantiations in LPC/MCX devices.

FelixWang47831 (changes requested):
Thank for your fix. I have verified the code and sctimer RM. This patch is correct. Just need some update:
1. move line 496 after line 505, argument validation should be performed before applying the settings.
2. About commit log "The User Manual is misleading here." is not clear. Please add explanation. 
For example:
"
HEVENT field must be set for all Counter_H events.
The User Manual is not accurate about HEVENT field description.
For  HEVENT field, when UNIFY = 0, binds this event to the L (0) or H (1) counter for three purposes: 
       1. STATEMSKn is compared against STATE_L/H; 
       2. MATCHSEL refers to the L/H half of the match register;
       3. any STATELD/STATEV update writes STATE_L/H.
"

einergehtnochrein:
I forgot to update the comment for my last commit, sorry. Your suggested explanation is accurate, and I would have added it unchanged...
My original comment about a misleading description in the manual was referring to the HEVENT description: "Selects the H state and the H match register selected by MATCHSEL". My wild guess is that this description, while technically correct, had suggested to the original driver author that HEVENT was relevant only for MATCH events.

FelixWang47831:
@einergehtnochrein You don't need to split 2 commit for this issue, just one commit is enough.
And please update the date to "2025-2026" in
https://github.com/einergehtnochrein/hal_nxp/blob/05b7e3753f2566d0003cf21bd3df261b506d641e/mcux/mcux-sdk-ng/drivers/sctimer/fsl_sctimer.c#L3

Update version to "2.5.4" in:
 https://github.com/einergehtnochrein/hal_nxp/blob/05b7e3753f2566d0003cf21bd3df261b506d641e/mcux/mcux-sdk-ng/drivers/sctimer/fsl_sctimer.h#L26
https://github.com/einergehtnochrein/hal_nxp/blob/05b7e3753f2566d0003cf21bd3df261b506d641e/mcux/mcux-sdk-ng/drivers/sctimer/CMakeLists.txt#L6

And changelog in https://github.com/einergehtnochrein/hal_nxp/blob/05b7e3753f2566d0003cf21bd3df261b506d641e/mcux/mcux-sdk-ng/drivers/sctimer/doxygen/ChangeLog_sctimer.md:
```
## [2.5.4]

- Bug Fixes
  - Fixed SCTIMER_CreateAndScheduleEvent setting EV[n].CTRL[HEVENT] only for
    match-based events.
```

Thanks.

PR #730, opened 2026-05-14
mcux-sdk-ng: components: Supports calculation of FW download time

xinyu0322:
Supports calculation of FW download time

### DIFF
--- a/mcux/mcux-sdk-ng/components/conn_fwloader/fsl_loader_utils.c
+++ b/mcux/mcux-sdk-ng/components/conn_fwloader/fsl_loader_utils.c
@@ -310,6 +310,8 @@ status_t sb3_fw_download_impl(LOAD_Target_Type loadTarget, uint32_t flag, uint32
         return status;
     }
 
+    PRINTF("[FW Download] Start to download %u firmware: %u\r\n", target_type, OSA_TimeGetMsec());
+
     if (loadTarget & 0x80)
     {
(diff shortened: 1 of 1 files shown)


MaochenWang1 (approved):

mcux/mcux-sdk-ng/components/conn_fwloader/fsl_loader_utils.c
```diff
@@ -310,6 +310,8 @@ status_t sb3_fw_download_impl(LOAD_Target_Type loadTarget, uint32_t flag, uint32
         return status;
     }
 
+    PRINTF("[FW Download] Start to download %u firmware: %u\r\n", target_type, OSA_TimeGetMsec());
```
zejiang0jason, line 313: MCUX SDK's debug console is not added in hal_nxp, how is this function used?

xinyu0322: The purpose of adding this log is to allow the firmware download time to be observed on the console. There is a customer concerned about the firmware download time, so QA requires this console log to record the current firmware download duration.

ZhaoxiangJin (approved):

MaochenWang1 (changes requested):
hold this change since `MCUX SDK's debug console is not added in hal_nxp`

PR #747, opened 2026-06-09
mcux: middleware: multicore: rpmsg-lite: restore Zephyr integration f…

TomasGalbickaNXP:
…iles

The 26.06.0-pvw2 multicore re-sync (commit 0fc95a88c) inadvertently regressed the Zephyr RPMsg-Lite integration that was originally added in commit 96c8c1c (Middleware multicore add Zephyr related files):

  * zephyr/Kconfig lost the 'RPMSG-Lite Configuration(rpmsg_config.h)' menu, so symbols such as RL_API_HAS_ZEROCOPY and RL_USE_STATIC_API became undefined. Assigning them in a prj.conf then aborts the build with 'attempt to assign the value to the undefined symbol' Kconfig warnings (treated as errors under sysbuild).

  * lib/include/platform/zephyr-ipc-service/rpmsg_config.h was rewritten to hardcode the SDK default values instead of reading the CONFIG_RL_* symbols, which made the options no-ops. The Kconfig-driven plumbing is restored here (each RL_* macro is derived from its CONFIG_RL_* counterpart) and the RL_USE_DCACHE bridge is added.

  * zephyr/CMakeLists.txt and lib/include/platform/zephyr-ipc-service/ rpmsg_platform.h are restored to the original Zephyr integration (VRING_ALIGN 256U, custom shmem config struct, RPMSGLITE_DIR fallback).

Restore the full set of Zephyr integration files so the IPC service RPMsg-Lite backend builds and honors its RL_* Kconfig options again.


ZhaoxiangJin (approved):
@yeaissa please take a look

PR #751, opened 2026-06-12
mcux: wifi_nxp: add IR OOB reset GPIO support for MCXN947

Liam-qk:
Add IR_OUTBAND_TRIGGER_GPIO definitions for MCXN947 (GPIO1, pin 22) when IW610 is used, following the same pattern as RT1060 and RT1170.

Also handle the gpio_pin_config_t initialization difference on MCXN947, which uses Kinetis-style GPIO (fsl_gpio) with a 2-field struct that does not include an interrupt mode field.


MaochenWang1:
please rebase this change on the TOT of master branch.

Liam-qk:
For wifi_nxp wireless SOC, no need to configure OOB GPIO in wifi driver.

PR #753, opened 2026-06-17
mcux: utilities: fsl_memcpy: add __aeabi_memcpy aliases

zejiang0jason:
Provide __aeabi_memcpy, __aeabi_memcpy4 and __aeabi_memcpy8 as aliases for the optimised memcpy implementation.

When Zephyr's LLEXT subsystem is enabled it exports these ARM EABI helpers via EXPORT_GROUP_SYMBOL so that loadable extensions can reference them.  Without local definitions the linker resolves the symbols from picolibc's memcpy.c.o, which also carries a strong `memcpy` symbol, causing a "multiple definition of `memcpy'" link error against fsl_memcpy.S.

With this change, the __aeabi_memcpy* in this file can be used.

### DIFF
--- a/mcux/mcux-sdk-ng/components/misc_utilities/fsl_memcpy.S
+++ b/mcux/mcux-sdk-ng/components/misc_utilities/fsl_memcpy.S
@@ -276,4 +276,11 @@ size_lt_4:                             /* size less than 4. */
 ret:
     pop    {r0, r4, r5, r6, r7, pc}
 
+    .global __aeabi_memcpy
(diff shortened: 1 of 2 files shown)


iuliana-prodan (approved):

zejiang0jason:
have submit PR in MCUX SDK.

ZhaoxiangJin (approved):

MaochenWang1:
what about fsl_memset.S?

MaochenWang1 (approved):

PR #754, opened 2026-06-19
hal_nxp: imx8mp: add LCDIFv3 compatibility layer

pwattebledsfl:
### Summary

Add a small HAL-layer compatibility shim for LCDIFv3 on i.MX8MP.

The i.MX8MP LCDIF register map is compatible with the existing single-layer LCDIFv3 HAL, but the generated i.MX8MP device headers name some registers differently from the i.MX93 headers that `fsl_lcdifv3.c` was written against.

For example, the LCDIFv3 HAL/driver expects accesses like:

```c
base->CTRL.RW
base->CTRLDESCL_LOW_4[0]
base->CTRLDESCL_5[0]
```

but the i.MX8MP generated header exposes those registers using the flat/layer-0 naming style, such as:
```c
CTRL
CTRLDESCL0_4
CTRLDESCL0_5
```

This PR adds `fsl_lcdifv3_soc.h`, which maps `LCDIF_Type` to an i.MX93-shaped compatibility layout on i.MX8MP. The hardware MMIO offsets do not change; only the C struct view used by the HAL changes. This lets existing LCDIFv3 code build unchanged while still accessing the correct i.MX8MP registers.

The shim also adds the matching bitfield macro aliases, for example:
```c
LCDIF_CTRLDESCL_5_BPP -> LCDIF_CTRLDESCL0_5_BPP
```

### Why this belongs in HAL
This is a mismatch between MCUX HAL source expectations and generated NXP device headers. Keeping the shim in HAL avoids modifying generated `PERI_LCDIF.h` headers and avoids adding any Zephyr-specific duplicate display driver logic.

### Notes
- This subject has been discussed in the Zephyr discord [Vendors-nxp channel](https://discord.com/channels/720317445772017664/883445396284772413/1516748589379096698)
- PERI_LCDIF.h is not modified.
- fsl_lcdifv3.c remains unchanged.
- The compatibility layer is guarded to i.MX8MP CPU macros only.
- Verified with an i.MX8MP/M7 DDR build using the LCDIFv3 Zephyr driver on a Waveshare 5' DSI TOUCH A display.

### DIFF
--- a/mcux/mcux-sdk-ng/drivers/lcdifv3/fsl_lcdifv3.h
+++ b/mcux/mcux-sdk-ng/drivers/lcdifv3/fsl_lcdifv3.h
@@ -8,6 +8,7 @@
 #define FSL_LCDIFV3_H_
 
 #include "fsl_common.h"
+#include "fsl_lcdifv3_soc.h"
 
 #if (defined(FSL_FEATURE_MEMORY_HAS_ADDRESS_OFFSET) && (0 != FSL_FEATURE_MEMORY_HAS_ADDRESS_OFFSET))
 #include "fsl_memory.h"
--- a/mcux/mcux-sdk-ng/drivers/lcdifv3/fsl_lcdifv3_soc.h
+++ b/mcux/mcux-sdk-ng/drivers/lcdifv3/fsl_lcdifv3_soc.h
@@ -1,0 +1,115 @@
+/*
+ * Copyright (C) 2026, Savoir-faire Linux
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef FSL_LCDIFV3_SOC_H_
+#define FSL_LCDIFV3_SOC_H_
+
+/* Per-SoC compatibility definitions for LCDIFv3. */
+
+#if defined(CPU_MIMX8ML2CVNKZ_cm7)  || defined(CPU_MIMX8ML2DVNLZ_cm7)  || \
+    defined(CPU_MIMX8ML2CVNKZ_ca53) || defined(CPU_MIMX8ML2DVNLZ_ca53) || \
+    defined(CPU_MIMX8ML2CVNKZ_dsp)  || defined(CPU_MIMX8ML2DVNLZ_dsp)  || \
+    defined(CPU_MIMX8ML3CVNKZ_cm7)  || defined(CPU_MIMX8ML3DVNLZ_cm7)  || \
+    defined(CPU_MIMX8ML3CVNKZ_ca53) || defined(CPU_MIMX8ML3DVNLZ_ca53) || \
+    defined(CPU_MIMX8ML3CVNKZ_dsp)  || defined(CPU_MIMX8ML3DVNLZ_dsp)  || \
+    defined(CPU_MIMX8ML4CVNKZ_cm7)  || defined(CPU_MIMX8ML4DVNLZ_cm7)  || \
+    defined(CPU_MIMX8ML4CVNKZ_ca53) || defined(CPU_MIMX8ML4DVNLZ_ca53) || \
+    defined(CPU_MIMX8ML5CVNKZ_cm7)  || defined(CPU_MIMX8ML5DVNLZ_cm7)  || \
+    defined(CPU_MIMX8ML5CVNKZ_ca53) || defined(CPU_MIMX8ML5DVNLZ_ca53) || \
+    defined(CPU_MIMX8ML5CVNKZ_dsp)  || defined(CPU_MIMX8ML5DVNLZ_dsp)  || \
+    defined(CPU_MIMX8ML6CVNKZ_cm7)  || defined(CPU_MIMX8ML6DVNLZ_cm7)  || \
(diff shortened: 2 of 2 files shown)


mmahadevan108:
@zejiang0jason @JiafeiPan Please review this PR.

zejiang0jason:
Hi @pwattebledsfl , thanks for the contribution, we are improving the LCDIFv3 driver to be compatible with these two types of headers.

zejiang0jason:
Hi @pwattebledsfl , could you please help check whether this meets your requirement? Thanks. I only test the compile, but not the function.

https://github.com/zephyrproject-rtos/hal_nxp/pull/785

pwattebledsfl:
> Hi @pwattebledsfl , could you please help check whether this meets your requirement? Thanks. I only test the compile, but not the function.
> 
> #785

Hi @zejiang0jason, thanks for looking into this. I compiled and tested my PR #754 on an i.MX8MP EVK with a real display, and it worked correctly.

I also tested your PR #785 against my Zephyr application. Its compatibility macros correctly bridge the i.MX8MP PERI_LCDIF.h register layout with fsl_lcdifv3.c and fsl_lcdifv3.h So both PRs compiles and works !

However, #785 does not cover consumers outside the HAL that directly access LCDIF_Type. In particular, Zephyr’s drivers/display/display_mcux_lcdifv3.c::dump_reg() still uses the i.MX93-style members CTRL.RW and CTRLDESCL_*[0], so it fails to compile with the i.MX8MP LCDIF_Type.

My PR avoided this by providing an i.MX93-shaped compatibility structure, making the HAL and its Zephyr consumer compatible without further changes. With the approach in #785, I think dump_reg() should instead use the new LCDIFV3_*_REG macros. This would preserve the generated LCDIF_Type while allowing the associated Zephyr driver to support both header formats.

I also encountered an undefined LCDIFV3_CLOCKS when compiling fsl_lcdifv3.c for MIMX8ML8, because its fsl_clock.h does not currently provide that mapping. To fix this issue, I added those lines under your `Format A` header:

```diff
+/* LCDIF1 is gated by the shared MEDIA CCGR on i.MX8MP. */
+#define LCDIFV3_CLOCKS { kCLOCK_Media }
```

zejiang0jason:
Hi @pwattebledsfl ，

> In particular, Zephyr’s drivers/display/display_mcux_lcdifv3.c::dump_reg() still uses the i.MX93-style members CTRL.RW and CTRLDESCL_*[0], so it fails to compile with the i.MX8MP LCDIF_Type

Thanks, I was not aware of that. Suggest update that too.

The `LCDIFV3_CLOCKS ` can be added too.
```c
#define LCDIFV3_CLOCKS { kCLOCK_Media, kCLOCK_Media, kCLOCK_Hdmi }
```

pwattebledsfl:
> Hi @pwattebledsfl ，
> 
> > In particular, Zephyr’s drivers/display/display_mcux_lcdifv3.c::dump_reg() still uses the i.MX93-style members CTRL.RW and CTRLDESCL_*[0], so it fails to compile with the i.MX8MP LCDIF_Type
> 
> Thanks, I was not aware of that. Suggest update that too.
> 
> The `LCDIFV3_CLOCKS ` can be added too.
> 
> ```c
> #define LCDIFV3_CLOCKS { kCLOCK_Media, kCLOCK_Media, kCLOCK_Hdmi }
> ```

Done, `LCDIFV3_CLOCKS` fix has been applied on this PR and tested on the imx8mp evk with the drivers/display/display_mcux_lcdifv3.c driver of zephyr and it works perfectly.

zejiang0jason:
Hi @pwattebledsfl , I will update the LCDIFv3 driver and LCDIFV3_CLOCKS in #785. and merge after our on-going 26.09 pvw2 sync finished. thanks.

mmahadevan108:
@zejiang0jason , please revisit this PR.

mmahadevan108:
@zejiang0jason please revisit this PR.

zejiang0jason:
Hi @mmahadevan108 , this PR will be replaced by https://github.com/zephyrproject-rtos/hal_nxp/pull/785.
Hi @pwattebledsfl could you please help update the description in https://github.com/zephyrproject-rtos/zephyr/pull/116345,  change the hal_nxp PR id to 785? thanks.

mmahadevan108:
@pwattebledsfl Please confirm that this approach works for you.
We could then close this PR.

jerome-lamy:
> @pwattebledsfl Please confirm that this approach works for you. We could then close this PR.

Hi @mmahadevan108 , @pwattebledsfl was doing an internship in our compagny on this driver. But unfortunately it ended,  I picked up the subject but don't have access to his account to modify the description. Does an administrator would be able to do it ?

mmahadevan108:
> > @pwattebledsfl Please confirm that this approach works for you. We could then close this PR.
> 
> Hi @mmahadevan108 , @pwattebledsfl was doing an internship in our compagny on this driver. But unfortunately it ended, I picked up the subject but don't have access to his account to modify the description. Does an administrator would be able to do it ?

This is something to check with github. We will go ahead and merge the alternate solution from @zejiang0jason and we can close this PR.

zejiang0jason:
use the solution in: https://github.com/zephyrproject-rtos/hal_nxp/pull/785

PR #756, opened 2026-06-25
device: RW: power: fix GDET disableCount corruption on PM3 entry failure

gangli02:
When PM3 entry fails (PMU->PWR_MODE_STATUS != 2), PostPowerMode() resets s_gdetSensorContext.disableCount to 0, since ROM never ran, GDET is still disabled from PrePowerMode().

This causes disableCount to underflow (0xFFFFFFFF) if POWER_EnableGDetVSensors() is called later (e.g., during Wi-Fi FW download), and subsequent POWER_DisableGDetVSensors() restores it to 0. The next PM2/PM3 entry then hits the assert in POWER_PrePowerMode() because disableCount == 0.


axelnxp (approved):
I assume this needs to be back ported into the mcuxsdk ?

gangli02:
> I assume this needs to be back ported into the mcuxsdk ?

yes, I have raised a PR to mcuxsdk.

MaochenWang1 (approved):

ZhaoxiangJin (approved):

zejiang0jason (approved):

PR #771, opened 2026-07-17
dts: nxp: pinctrl deduplication into wrappers and delta wrappers

lucien-nxp:
## Motivation

Customers report hal_nxp is too large. `dts/nxp` alone carried 64MB of generated pinctrl files, one per orderable part number, although parts of the same die and package are byte-identical (temperature/speed grades) or differ only in a handful of signals.

## What this PR does

**Commit 1 - tooling** (`mcux/scripts/pinctrl/`):
- `dedup_pinctrl.py`: clusters pinctrl files by content and rewrites duplicates as one-line wrappers that `#include` the cluster's canonical file (the highest-numbered flagship part). Parts that differ only partially become *delta wrappers*: `.h` files include the base, `#undef` the signals the part lacks and define the differing/additional ones; `.dtsi` files include the base, `/delete-node/` the differing nodes and re-add their own. Within equal part numbers the file with more entries is preferred as the base — a die has one full mux set and derived parts are subsets, so most deltas are pure removals. The effective preprocessor / devicetree content of every part number is unchanged.
- Built-in verification: every `--apply` run verifies each rewritten file against its original content on the spot (`gcc -E -dM` effective-macro equality for `.h`, dtlib devicetree equality for `.dtsi`); failures are reverted and the run exits non-zero. Every wrapper records a `target-hash` of the content it was generated against and stale wrappers fail the run, so regenerating only a base part cannot silently change its dependents.
- `gen_soc_headers.py` now deduplicates freshly generated files by default, including against files already present in the output directory, so a single-NPI data pack reuses existing identical data automatically (`--no-dedup` opts out).

**Commit 2 - generated result** (`dts/nxp`): 331 exact wrappers + 241 delta wrappers, shrinking `dts/nxp` from 64MB to 14MB (-78%). Every part number file keeps existing under its original name, so board devicetree includes are unaffected.

## Verification

All 572 rewritten files were verified twice: in-run by the tool itself, and cross-checked against the pre-change tree with an independent script — `gcc -E -dM` macro set and value equality for `.h` (399), dtlib devicetree semantic equality for `.dtsi` (149), normalized content equality for 24 LPC54xxx files whose originals gcc cannot parse (pre-existing invalid `SPIFI_IO(0)` macros from the generator). Changed-file set matches the plan exactly. A companion zephyr manifest PR runs the full CI/twister against this revision.

## Notes for reviewers

- Wrapper include guards are derived from the wrapper's own file name: some generated files ship with another part's guard (e.g. MCXW235BIHNAR carries the MCXW235AIHNAR guard), which must not be reused.
- Found during verification, worth fixing in the generator upstream: invalid macro names in LPC54xxx (`SPIFI_IO(0)_PIO0_24`) and MCXA (`TRIG_IN5/CT_INP19_...`) headers, and cross-part include guards (MCXW235B).


lucien-nxp:
Add @peterwangsz @NeilChen93 @ZhaoQiang-b45475 @JiafeiPan @yangbolu1991 @biwenli @Dat-NguyenDuy @haduongquang @manuargue  for notice this change. 
Any comment is welcome, thank you.

biwenli:
It's okay from my side.

peterwangsz:
it is OK for MCX pinctrl update.

mcux/scripts/pinctrl/gen_soc_headers.py
```diff
@@ -202,6 +211,34 @@ def main():
...
+    if not args.no_dedup and written_files:
+        # Deduplicate the generated files against each other AND against
+        # the pinctrl files already present in the output directories, so
+        # a single-NPI data pack can reuse an existing part's data (or
+        # become the new canonical if it is the higher-numbered flagship).
+        dedup_set = set()
+        for out_file in written_files:
+            dedup_set.add(pathlib.Path(out_file))
+            dedup_set.update(pathlib.Path(out_file).parent.glob("*-pinctrl.*"))
+        clusters, deltas, stale, failures = dedup_pinctrl.dedup_files(
```
hakehuang, line 223: CRASH BUG — dedup_files() return/unpack mismatch dedup_files() in dedup_pinctrl.py returns 3 values:

return clusters, deltas, stale   # line 292
But gen_soc_headers.py unpacks 4:

clusters, deltas, stale, failures = dedup_pinctrl.dedup_files(...)
This is a ValueError crash on every gen_soc_headers.py run that triggers dedup. Fix: add failures to the dedup_files() return:

return clusters, deltas, stale, failures

mcux/scripts/pinctrl/README.md
```diff
@@ -122,3 +123,75 @@ the `--controller` option in order to attempt to generate pin control data:
...
+
+`gen_soc_headers.py` applies the same deduplication to freshly generated
+files by default, so new SOC data packs produce deduplicated output
+directly; pass `--no-dedup` to keep the full per-part files. The
+generated files are also clustered against the pinctrl files already
+present in the output directory, so a single-NPI data pack reuses an
+existing part's identical data automatically. If the new part is the
+higher-numbered flagship it takes over as the canonical file: the old
+canonical becomes a wrapper and all existing wrappers are re-pointed to
+the new canonical, so every wrapper stays exactly one include away.
```
hakehuang: 2-level #include chain in committed output — README guarantee violated K32L2A31VLH1A-pinctrl.h (exact wrapper) → K32L2A41VLH1A-pinctrl.h (delta wrapper) → K32L2A41VLL1A-pinctrl.h (canonical)

flatten_wrappers() explicitly stops at delta wrappers (if next_kind != "exact": break), so the exact wrapper is left pointing at the delta, creating a 2-level chain. The README states:

"every wrapper stays exactly one include away"

That guarantee is broken in the committed files.

mcux/scripts/pinctrl/gen_soc_headers.py
lucien-nxp: Good catch — failures was computed inside dedup_files() but never returned. Fixed as suggested: it now returns (clusters, deltas, stale, failures).

mcux/scripts/pinctrl/README.md
lucien-nxp: This is a description error here. The chain is real but necessary: K32L2A31VLH1A equals the delta result, not the canonical, so re-pointing it past the delta would change its content. Avoiding the hop would mean duplicating the delta body, costing the bytes the dedup saves. flatten_wrappers() therefore only flattens through exact wrappers by design. 
Fixed the README wording instead — it no longer claims every wrapper is one include away.

ZhaoQiang-b45475:
Looks good to me.

hakehuang (approved):

lucien-nxp:
Hi @ZhaoxiangJin 
Could you help check and help merge the PR? Test on zephyr side had been tested passed. Thank you!

ZhaoxiangJin (approved):

PR #772, opened 2026-07-20
wifi_nxp: wlcmgr: add missing stdlib.h includes

zoltangyarmati:
atoi() and strtol() are used in wlan_tests.c, wlan_enhanced_tests.c,
wlan_features.c, and wlan_test_mode_tests.c without including <stdlib.h>. Normally it's transitively included when built with NXP SDK, but they break the compilation when building with picolibc


MaochenWang1:
Hi @zoltangyarmati 
what's your build cmd and config file? I didn't see this issue when build wifi shell example:
`west build -p always -b rd_rw612_bga samples/net/wifi/shell -d wifi`

PR #784, opened 2026-08-05
cmake: Add check for existing flag to avoid accumulation across cmake runs

Larsjep:
Without this check the list of flags will accumulate across cmake runs, which causes a full rebuild because these are global flags.

Fixes Zephyr issue: https://github.com/zephyrproject-rtos/zephyr/issues/115425

### DIFF
--- a/mcux/mcux-sdk-ng/cmake/extension/function.cmake
+++ b/mcux/mcux-sdk-ng/cmake/extension/function.cmake
@@ -2143,6 +2143,13 @@ function(mcux_append_single_compiler_flags_variable type flag)
      mcux_set_property(MCUX_${type}_MACRO_${CMAKE_MATCH_1} ${MACRO_VALUE})
    endif()
  endif()
+
+  # Check if the flag already exists to avoid accumulation across cmake runs
+  string(FIND " ${CMAKE_${lang_type}_FLAGS} " " ${flag} " _flag_pos)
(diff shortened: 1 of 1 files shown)


zejiang0jason:
Hi @Holt-Sun , could you please review this? thanks.

Holt-Sun:
A few call sites pass flags with spaces as a single arg, e.g. in `mcux_add_sys_include`:

    mcux_append_single_compiler_flags_variable(CC "-I- -i ${item}")

Since the guard splits existing flags on spaces, `"-I- -i /path"` won't match as one item and still accumulates. These are IAR flags so the rw612/GCC repro is unaffected, but the dedup no-ops for them.

Maybe match against the whole string instead of a split list? `string(FIND ...)` would cover both cases (and dodges regex-escaping the flag).

mmahadevan108:
@Larsjep , can you please revisit this PR based on feedback from Holt.

Larsjep:
> @Larsjep , can you please revisit this PR based on feedback from Holt.

Yes, I will do that.

Larsjep:
Hi @mmahadevan108 and @Holt-Sun 

Sorry for the delay, but now it should also work for flags with spaces.

/Lars

mmahadevan108:
@Holt-Sun did you get a chance to look at this?

PR #795, opened 2026-08-20
wifi_nxp: support enhance command when enable NXP_WIFI_RF_TEST_MODE

GaofengZhangNXP:
support enhance command when enable NXP_WIFI_RF_TEST_MODE


MaochenWang1:
already merged in downstream, this one is not needed any more

PR #796, opened 2026-08-20
dts: nxp: restructure pinctrl into family/series layout

lucien-nxp:
Restructure the SoC pin control headers under `dts/nxp` to mirror the
zephyr DTS side two-level `dts/arm/nxp/<family>/<series>/` layout: every
SoC `*-pinctrl.h` / `*-pinctrl.dtsi` file moves into a
`<family>/<series>/pinctrl/` subdirectory. A sibling `clock/` directory is
intentionally reserved for a future change and is **not** created here.

### Layout rules

- The families that are series-organized on the DTS side (imxrt, kinetis,
  lpc, mcx) get a `<series>/pinctrl/` subdirectory.
- The families that are flat on the DTS side (imx, s32, rw) keep their
  headers at the family root.
- kinetis is series-organized, so kinetis parts always move into a series
  dir, creating `k0x`/`km3x`/`kv3x` for the parts that had no matching
  series dir yet.
- The `nxp_imx` directory is split: `nxp_imx/rt/*` becomes the `imxrt/`
  family and the remaining i.MX application processors become the flat
  `imx/` family.

### Generator

`gen_soc_headers.py` is updated so regeneration writes into the new layout
instead of re-flattening the tree. A new module,
`mcux/scripts/pinctrl/shared/pinctrl_layout.py`, is the single source of
truth mapping a part number to its `<family>/<series>/pinctrl` sub-dir
(DTS-side authority + kinetis exception + flat fallback), and was validated
against all 661 placed files (0 mismatch). The pinctrl README documents the
`--soc-output` convention.

### Notes

- Signal file moves are pure renames (git rename detection reports R100); no
  pinctrl content changes.
- Requires a coordinated zephyr change to update the in-tree
  `#include <nxp/...-pinctrl.*>` paths; that PR will pin the west manifest to
  this PR.

Signed-off-by: Lucien Zhao <lucien.zhao@nxp.com>


ZhaoxiangJin:
`pinctrl` in `dts/nxp/imx/pinctrl/<soc>-pinctrl.dtsi‎` path is not needed I think.

lucien-nxp:
> `pinctrl` in `dts/nxp/imx/pinctrl/<soc>-pinctrl.dtsi‎` path is not needed I think.

This lays the foundation for migrating SoC-specific clock parameters into Devicetree in the future. Some SoC-specific clock parameters required by Zephyr DTS will be defined under the corresponding SoC series directories in hal_nxp/dts. To support this, a dedicated "clock" directory will be introduced for clock-related SoC description files. This directory is not intended to be used solely for pinctrl data storage. Going forward, it can also host other SoC-specific content that is outside the Zephyr generic layer.

I have already completed the clock driver implementation for the new product. However, due to schedule constraints, I am not planning to upstream it for now.

zejiang0jason (approved):

PR #798, opened 2026-08-20
drivers: clock: rw6xx: fix wrong ENET MDC clock frequency

Varghesemela:
CLOCK_GetTddrMciEnetClkFreq() hardcodes 50MHz with no register read. This value is used by ENET_SetSMI() to calculate the ENET module's MSCR MII_SPEED/HOLDTIME divider, on the assumption that it reflects the module's real clock source (TDDR).

In reality, on RW610/RW612 that divider is calculated against the bus clock, which runs at 260MHz, not 50MHz. With the wrong assumed base clock, the resulting real MDC frequency comes out around 13MHz instead of the intended 2.5MHz, exceeding common RMII PHY MIIM interface limits (e.g. KSZ8081, ~10MHz max) and causing unreliable or completely non-functional MDIO.

I have reproduced this on my custom RW610 board, and was independently confirmed by NXP (Pavel Krenek) on an old community forum discussion, including the same fix: [Link to NXP forum discussion](https://community.nxp.com/t5/Zephyr-Project/Wrong-MDC-frequency-RW612-frdm-rw612/td-p/2054592)

Fix both the RW610 and RW612 copies of this stub to report the real 260MHz clock, so ENET_SetSMI() computes a correct MII_SPEED/HOLDTIME for actual MDC operation.

### DIFF
--- a/mcux/mcux-sdk-ng/devices/Wireless/RW/RW610/drivers/fsl_clock.c
+++ b/mcux/mcux-sdk-ng/devices/Wireless/RW/RW610/drivers/fsl_clock.c
@@ -236,7 +236,15 @@ uint32_t CLOCK_GetTddrMciFlexspiClkFreq(void)
  */
 uint32_t CLOCK_GetTddrMciEnetClkFreq(void)
 {
-    uint32_t freq = CLOCK_MHZ(50UL);
+    /* The ENET module's MSCR/MDC divider is calculated by ENET_SetSMI() against
+     * whatever this function reports, but it is actually clocked from the M33
+     * processor/platform bus clock (260MHz), not from TDDR. The previous 50MHz
+     * value produced a real MDC of ~13MHz instead of the intended 2.5MHz, which
+     * exceeds the KSZ8081 (and most RMII PHYs') MIIM interface limits and causes
+     * unreliable/non-functional MDIO. Confirmed root cause and fix:
+     * https://community.nxp.com/t5/Zephyr-Project/Wrong-MDC-frequency-RW612-frdm-rw612/td-p/2054592
+     */
+    uint32_t freq = CLOCK_MHZ(260UL);
     return freq;
 }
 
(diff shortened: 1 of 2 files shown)


zejiang0jason:
@AllenZhang-NXP is checking and picking to SDK if reasonable.

AllenZhang-NXP:
Thanks for the detailed report and for tracking down the MDC frequency problem — the symptom you're seeing on RW610/RW612 is real. However, after reviewing this with the NXP ENET driver owner, I don't think changing CLOCK_GetTddrMciEnetClkFreq() is the right fix, because that function is intentionally the PTP reference clock, not the SMI/MDC base clock.

Why the shared stub should stay at 50 MHz
In the MCUXpresso SDK the SMI/MDC path and the PTP path are already split and use two different clock accessors:

/* examples/.../enet/txrx_ptp1588_transfer/app.h */
#define PTP_CLK_FREQ       CLOCK_GetTddrMciEnetClkFreq()   /* PTP timestamp clock only */
#define EXAMPLE_CLOCK_FREQ CLOCK_GetMainClkFreq()          /* base clock for SMI/MDC   */

/* hardware_init.c */
ENET_SetSMI(EXAMPLE_ENET, EXAMPLE_CLOCK_FREQ, false);      /* uses main clock, NOT TddrMciEnet */

So in the SDK, CLOCK_GetTddrMciEnetClkFreq() (50 MHz) is never used to compute the MDC divider — ENET_SetSMI() is fed CLOCK_GetMainClkFreq() instead. Its 50 MHz return value is the PTP timestamp time base, and changing it to 260 MHz would break PTP timestamping.

Where the RW6xx bug actually is (Zephyr side)

The wrong MDC comes from the Zephyr clock-control driver, which returns this PTP stub as the generic ENET clock that the Zephyr ENET driver then uses for the SMI divider:

/* drivers/clock_control/clock_control_mcux_syscon.c */
case MCUX_ENET_CLK:
#ifdef CONFIG_SOC_SERIES_RW6XX
    *rate = CLOCK_GetTddrMciEnetClkFreq();   /* <-- returns 50 MHz, used for MDC divider -> ~13 MHz MDC */
#endif
    break;

The correct fix is to make MCUX_ENET_CLK on RW6xx report the actual bus/main clock, the same value the SDK feeds to ENET_SetSMI(), rather than the PTP stub. Note the MCUX_ENET_QOS_CLK case just above already does the right thing with CLOCK_GetFreq(kCLOCK_BusClk).

Varghesemela:
Thanks for looking into this @AllenZhang-NXP 
Are you planning to implement the proposed fix, or I can work on this if it seems straighforward.

PR #807, opened 2026-09-03
component: update rb3+ compress tx power limit table

sanjay-yadav-nxp:
update rb3+ compress tx power limit table for AzureWave


mmahadevan108:
Where are we getting these new numbers from?

sanjay-yadav-nxp:
> Where are we getting these new numbers from?

AzureWave

PR #808, opened 2026-09-03
drivers: wifi: enable compressed TX power table for IW416 and IW612

GaofengZhangNXP:
NXP_WIFI_COMPRESS_TX_PWTBL was not selectable for IW416, so IW416 kept using the legacy per-channel TRPC tx power tables. Add NXP_IW416 to the if block and to its depends on, and give NXP_WIFI_CLOCKSYNC an explicit depends on so it stays disabled for IW416.

Also drop the NXP_IW416_REGION and NXP_IW61X_REGION choices. The driver now selects the region table at runtime from the country code, and one WW header already carries the WW, JP, US, EU and CA tables, so the per-region headers are redundant. The header is selected per chip in the wifi_nxp CMakeLists instead, as RW610 already does.


mcux/middleware/wifi_nxp/CMakeLists.txt
```diff
@@ -34,10 +34,12 @@ zephyr_compile_definitions_ifdef(CONFIG_NXP_88W8987
 
 zephyr_compile_definitions_ifdef(CONFIG_NXP_IW416
     SD8978
+    WIFI_BT_TX_PWR_LIMITS="wlan_txpwrlimit_cfg_murata_1XK_WW.h"
 )
 
 zephyr_compile_definitions_ifdef(CONFIG_NXP_IW61X
     SD9177
+    WIFI_BT_TX_PWR_LIMITS="wlan_txpwrlimit_cfg_murata_2EL_WW.h"
```
sanjay-yadav-nxp: Please add logic such that it first checks the board typrthan based on it select the respective tx pwr file. As for IW61X we have SoCs from Murata, u-blox, QUECTEL etc. Same for IW416.

MaochenWang1 (approved):

sanjay-yadav-nxp (approved):

MaochenWang1:
Hi @mmahadevan108 @zejiang0jason please review and merge

zejiang0jason (approved):

zejiang0jason (approved):

PR #810, opened 2026-09-07
wifi_bt_module: murata: enable 40MHz on 2.4GHz band edge channels for 2EL

MaochenWang1:
The compressed TX power tables for Murata 2EL (JP, US, EU, CA regions)
had 40MHz power index values set to 0 for 2.4GHz band edge channels
(ch1, ch2, ch10, ch11). The rgpower_gen tool treats 0 values as
disabled, causing these channels to be flagged with 40MHz_DISABLED
in chanAttr2G. This prevents uAP from starting in 40MHz mode on
these channels.

Regenerated compressed TX power tables for JP, US, EU and CA regions
with 40MHz power values populated for band edge channels using the
corresponding center channel values. FW always applies center channel
power values for 40MHz operation, so the actual power values for
edge channels are unused, but non-zero values prevent the tool from
setting the 40MHz_DISABLED flag.


MaochenWang1:
This change is already merged in downstream mcuxsdk.

zejiang0jason (approved):

PR #814, opened 2026-09-10
nxp_crypto_hal: add ELS_PKC

VitSolcikNXP:
Add ELS_PKC cryptolib


VitSolcikNXP:
@mmahadevan108 @zejiang0jason Can you please check this PR? 
Top PR is [here](https://github.com/zephyrproject-rtos/zephyr/pull/118311)

zejiang0jason:
The header files in folder `includes` is a copy of the headers from other places, for example: `includes/platform/rw61x/ip_platform.h` is the same with `src/platforms/rw61x/inc/ip_platform.h` is this intentional?

VitSolcikNXP:
Unfortunately yes, some build configurations appears to link the common header files files from src path but the platform ones from "includes" because they are grouped together in one folder, which is beneficial for library builds.

EDIT: I'll try to dig deeper if we can rework the build infrastructure...
EDIT2: I was able to remove all files in the `includes/` folder and rework the Cmake. Build passed locally.

zejiang0jason (approved):

ZhaoxiangJin (approved):

PR #822, opened 2026-09-19
mcux: sdk-ng: prune unused mcux drivers

ZhaoxiangJin:
prune unused mcux drivers


henrikbrixandersen:
Have you considered that removing these prevents anybody from using them for out-of-tree shims?

ZhaoxiangJin:
> Have you considered that removing these prevents anybody from using them for out-of-tree shims?

Thank you for the review. This is a reasonable concern, and it was exactly the same reasoning that led us to narrow the scope of this PR: originally I had also considered removing 111 driver directories that have no consumer in Zephyr, but held back because an out-of-tree shim driver might depend on them.

The two commits in this PR go in different directions:

1. [Commit 1](https://github.com/zephyrproject-rtos/hal_nxp/pull/822/changes/a1c041342d1b30dc403b5183b43e7c78c651285a) removes 61 drivers. These drivers cannot be compiled in this tree, so they do not represent a usable out-of-tree starting point:

- `dsc_*` targets the `DSP56800EX` core. That architecture is absent from `arch/`, and the Zephyr SDK does not ship a DSC compiler (CodeWarrior is required to build it).
- `jn_*` (JN5189) and `qn_*` (QN9090) are Cortex-M4, so the architecture is not the blocker — but their device headers are missing from hal_nxp. mcux/mcux-sdk-ng/devices/ only covers `i.MX` / `Kinetis` / `LPC` / `MCX` / `RT` / `Wireless` (K32W, KW, RW) / `xtensa`; there are no `JN518x` or `QN9xxx` entries anywhere in the tree.
- Every `fsl_*.c` resolves through `fsl_common.h` → `fsl_device_registers.h`, which cannot resolve for these devices.
In other words, writing an out-of-tree shim for any of these devices would require restoring the full device header set first; recovering the two driver files would be the smallest step in that effort.

2. [Commit 2](https://github.com/zephyrproject-rtos/hal_nxp/pull/822/changes/8794bf10aaa7763a8948d8911c21b77c6cbc7e0f) removes 21 drivers. Your concern does apply to this commit. All 21 IPs exist on SoCs that Zephyr already supports. Although Zephyr-native drivers cover those peripherals, some IP-specific features are not exposed through the Zephyr device driver API, and an out-of-tree project might call the HAL driver API directly.

I will drop the second commit to avoid any regression for out-of-tree projects.

ZhaoxiangJin:
update:
rebased to fix conflicts

iuliana-prodan:
@ZhaoxiangJin Thanks for working on this.
Could you please open a Zephyr PR that updates `west.yml` and points to this PR so we can validate it through CI?

@hakehuang could you also take a look at this PR and run the regression tests, or any other test suite you consider?

ZhaoxiangJin:
> @ZhaoxiangJin Thanks for working on this. Could you please open a Zephyr PR that updates `west.yml` and points to this PR so we can validate it through CI?

already created, see https://github.com/zephyrproject-rtos/zephyr/pull/119653

> @hakehuang could you also take a look at this PR and run the regression tests, or any other test suite you consider?

Hake already help tested, result is good.

iuliana-prodan:
> > @ZhaoxiangJin Thanks for working on this. Could you please open a Zephyr PR that updates `west.yml` and points to this PR so we can validate it through CI?
> 
> already created, see [zephyrproject-rtos/zephyr#119653](https://github.com/zephyrproject-rtos/zephyr/pull/119653)
> 

Sorry, I've missed that.
Is approved now :)

> > @hakehuang could you also take a look at this PR and run the regression tests, or any other test suite you consider?
> 
> Hake already help tested, result is good.

Great.
Thank you!

iuliana-prodan (approved):

hakehuang (approved):

